PR #9, opened 2020-06-30
Implement Airdrop

ChihChengLiang:


contracts/airdrop.sol
```diff
@@ -0,0 +1,102 @@
...
+        );
+        uint256 balance = RollupUtils.BalanceFromAccount(new_account);
+        return (newRoot, balance);
+    }
+}
+
+contract Airdrop is AirdropHelpers {
+
+    function processDrop(
+        Types.Transaction memory drop,
```
ChihChengLiang, line 64: We parse the same transaction format but we care only the `to` and `amount` field only

contracts/airdrop.sol
```diff
@@ -0,0 +1,102 @@
...
+        Types.AccountMerkleProof memory _to_merkle_proof
+    )
+        public
+        view
+        returns (
+            bytes32,
+            uint256,
+            bool
+        )
+    {
```
ChihChengLiang, line 74: TODO: I think we still need to check the public key is available in the PDA tree.

contracts/libs/Types.sol
```diff
@@ -21,6 +21,7 @@ library Types {
         uint256 stakeCommitted;
         uint256 finalisesOn;
         uint256 timestamp;
+        BatchType batchType;
```
ChihChengLiang, line 24: We add BatchType for batches.

contracts/rollup.sol
```diff
@@ -361,17 +368,28 @@ contract Rollup is RollupHelpers {
         // start with false state
         bool isDisputeValid = false;
 
+        if (batches[_batch_id - 1].batchType == Types.BatchType.Airdrop) {
+            // Validate Reddit signature and sub reddit approval
+        }
```
ChihChengLiang, line 373: Only airdrop needs batch level checks, but not monthly consent and execution. Not sure if other future features need batch level checks.

contracts/rollup.sol
```diff
@@ -361,17 +368,28 @@ contract Rollup is RollupHelpers {
...
             // tx evaluates correctly
-            (newBalanceRoot, fromBalance, toBalance, isTxValid) = processTx(
-                batches[_batch_id - 1].stateRoot,
-                batches[_batch_id - 1].accountRoot,
-                _txs[i],
-                _pda_proof[i],
-                _from_proofs[i],
-                _to_proofs[i]
-            );
+            if (batches[_batch_id - 1].batchType == Types.BatchType.Transfer){
```
ChihChengLiang, line 378: We match BatchType and for each type, we run their process<Item> respectively.

contracts/libs/Types.sol
vaibhavchellani: Yeah that's a good point

contracts/rollup.sol
```diff
@@ -243,15 +247,18 @@ contract Rollup is RollupHelpers {
         fraudProof = IFraudProof(
             nameRegistry.getContractDetails(ParamManager.FRAUD_PROOF())
         );
-        addNewBatch(ZERO_BYTES32, genesisStateRoot);
+        airdrop = IAirdrop(
+            nameRegistry.getContractDetails(ParamManager.AIRDROP())
+        );
+        addNewBatch(ZERO_BYTES32, genesisStateRoot, Types.BatchType.Transfer);
```
vaibhavchellani, line 253: Maybe have another batch type called `genesis` and use that here.

contracts/libs/Types.sol
vaibhavchellani: Is the idea here to have a batchType and its `slashable` if there is a tx in that batch that doesnt belong there?

contracts/rollup.sol
vaibhavchellani: Yep, this is great!

contracts/airdrop.sol
vaibhavchellani: Yeah I think so too.

vaibhavchellani: Maybe add a TODO comment for signature check as well.

contracts/airdrop.sol
barryWhiteHat: One optimization for later is to have a differnet transatino format so that we don't need to put from on chain.

contracts/libs/Types.sol
ChihChengLiang: Reposting the conclusion from the offline discussion here. We don't need to check if the tx type matches the batch type explicitly. The sender signs the uncompressed tx with the tx type in it, so when we validate the signature of a compressed tx in the batch, the signature is only valid if the tx type is correct.

contracts/airdrop.sol
ChihChengLiang: For airdrop, we might not want to check the signature for each drop.

barryWhiteHat: why not?

ChihChengLiang: Ah, I mean we want to check 1 signature at the batch level, but don't want each transaction in that batch to have their own signature and check them many times. Does that make sense?

barryWhiteHat: I think we need to have one signature per tx. Because reddit will just sign a single message per user and then the user gives it to the coordinator.

ChihChengLiang:
Close in favor of #11

PR #11, opened 2020-07-02
Implement Airdrop 2

ChihChengLiang:
This time we implement on the new applyTx abstraction.


ChihChengLiang:
humm need some trick to reuse an interface https://ethereum.stackexchange.com/a/65288/1006

PR #17, opened 2020-07-09
Process drop

ChihChengLiang:


test/Airdrop.spec.ts
```diff
@@ -1,15 +1,82 @@
...
+const RollupUtils = artifacts.require("RollupUtils");
+const IMT = artifacts.require("IncrementalTree");
 
 contract("airdrop", async function () {
 
-    it("lets coordinator submit a batch", async function () {
+    const tokenType = 1
+    const epoch = 1
+
+    xit("lets coordinator submit a batch", async function () {
```
ChihChengLiang, line 13: I hesitate to give testing for even the validity path right now. The iteration of writing tests is a little bit slow.
 
Reasons:
- Every time you rerun a single test, it needs two minutes to redeploy every contract.
- Struct doesn't work specifically in tests but not in go-node or remix. See  https://github.com/thehubbleproject/contracts/issues/57. We have to rewrite the rollup utils format just to let the tests pass.
- It's difficult to create objects like Merkle proof or siblings manually. This suggests we need more utils or domain objects

scripts/helpers/setup.ts
```diff
@@ -0,0 +1,143 @@
...
+        Address: wallets[1].getAddressString(),
+        Pubkey: wallets[1].getPublicKeyString(),
+        Amount: 10,
+        TokenType: 1,
+        AccID: 2,
+        Path: "3",
+    };
+
+    await testTokenInstance.transfer(Alice.Address, 100);
+    await depositManagerInstance.deposit(
```
ChihChengLiang, line 56: These are written before we had the create account discussion.

PR #20, opened 2020-07-09
Use struct in process tx

ChihChengLiang:


vaibhavchellani:
Actually maybe a couple of other things are needed here:
1. Use  `Types.BurnConsent memory _tx` here as well: https://github.com/thehubbleproject/RedditHubble/blob/874a7a552f341630947673f5d1b192dc3e30b0cc/contracts/BurnConsent.sol#L40
2. Then use https://github.com/thehubbleproject/contracts/blob/master/contracts/FraudProof.sol#L244 to compress and create txRoot on-chain so you don't ever have to call decompress on tx on-chain.
3. You can then remove this line: https://github.com/thehubbleproject/RedditHubble/blob/874a7a552f341630947673f5d1b192dc3e30b0cc/contracts/BurnConsent.sol#L68

The bytes submitted by the user while calling fraud proof will be of `uncompressedTx`

PR #27, opened 2020-07-12
Merging upstream 4

ChihChengLiang:


contracts/rollup.sol
```diff
@@ -449,43 +447,14 @@ contract Rollup is RollupHelpers {
             bool
         )
     {
-        if (batchType == Types.Usage.Transfer) {
```
ChihChengLiang, line 452: @vaibhavchellani To merge the upstream we need to abort this part. But it's ok to do so since we are not using processBatch?

vaibhavchellani: We can do both. Lets quickly remove this and merge for now and later I can add this to upstream?

vaibhavchellani:
Also @ChihChengLiang can you make a pull of upstream again, I just merged something

contracts/rollup.sol
ChihChengLiang: Sounds good

PR #32, opened 2020-07-13
Refactor Utils

vaibhavchellani:
There are a lot of utils that need to be there for each transaction type so I figure its a good idea to bunch utils for each of them together.


vaibhavchellani:
Yeah I will make relevant changes in upstream too.

PR #36, opened 2020-07-16
Tests and bench CreateAccount

ChihChengLiang:


contracts/CreateAccount.sol
```diff
@@ -139,15 +139,15 @@ contract CreateAccount is FraudProofHelpers {
         bytes32 _accountsRoot,
         Types.CreateAccount memory _tx,
         Types.PDAMerkleProof memory _to_pda_proof,
-        Types.AccountProofs memory accountProofs
+        Types.AccountMerkleProof memory to_account_proof
```
ChihChengLiang, line 142: @vaibhavchellani No need passing `from` proofs here, so I removed the `from` proof and keep the `to` proof. Does this affect the node?

contracts/CreateAccount.sol
```diff
@@ -160,22 +160,38 @@ contract CreateAccount is FraudProofHelpers {
...
-        if (result == false) {
-            return ("", "", "", 0, false);
+        // Validate we are creating on a zero account
+        if (
+            !merkleUtils.verifyLeaf(
+                _balanceRoot,
+                keccak256(abi.encode(0)), // default hash 0
+                to_account_proof.accountIP.pathToAccount,
+                to_account_proof.siblings
+            )
```
ChihChengLiang, line 170: There's no actual zero account that looks like`{fromIndex: 0, toIndex:0 ...}`, we only have zero account leaf. So here's how we validate the proof.

test/Reddit.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+        const NewAccountSiblings: Array<string> = [
+            RedditAccountLeaf,
+            utils.getParentLeaf(coordinator_leaves[0], coordinator_leaves[1]),
+            zeroHashes[2],
+            zeroHashes[3],
+        ];
+        const NewAccountMP = {
+            accountIP: {
+                pathToAccount: User.Path,
+                account: ZeroAccount,
```
ChihChengLiang, line 91: The account field here can be an arbitrary account. We don't validate it since we validate zero account "leaf" only. We just need to pass something here because the input struct requires it.

test/Reddit.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+            zeroHashes[3],
+        ];
+        const userPDAProof = {
+            _pda: {
+                pathToPubkey: "3",
+                pubkey_leaf: { pubkey: User.Pubkey },
+            },
+            siblings: userPDAsiblings,
+        }
+        await createAccountInstance.processTx(
```
ChihChengLiang, line 111: Not sure why the gas reporter does not show the gas usage of this call.

contracts/CreateAccount.sol
vaibhavchellani: Nah, should be okay. No problem.

contracts/CreateAccount.sol
vaibhavchellani: Sounds good, just for the sake of saving some gas and not needing to hardcode things, can you use [this](https://github.com/thehubbleproject/contracts/blob/master/contracts/MerkleTreeUtils.sol#L9) instead of `keccak256(abi.encode(0))`?

vaibhavchellani: So just merkleUtils.defaulutHashes(0) should work

test/Reddit.spec.ts
vaibhavchellani: We can just call estimate_call on this and get the gas for now I think.

contracts/CreateAccount.sol
```diff
@@ -160,22 +160,38 @@ contract CreateAccount is FraudProofHelpers {
         // Assuming Reddit have run rollup.sol::createPublickeys
         ValidatePubkeyAvailability(_accountsRoot, _to_pda_proof, _tx.toIndex);
 
-        // accountProofs.to.accountIP.account should to be a zero account
-        bool result = ValidateZeroAccount(accountProofs.to.accountIP.account);
-        if (result == false) {
-            return ("", "", "", 0, false);
+        // Validate we are creating on a zero account
```
vaibhavchellani, line 163: We need a signature check too I think?

ChihChengLiang: This requires passing "from" PDA proof to processTx, and we'll have 2 PDA proofs (from and to) in the processTx. If we are doing BLS and validate this at the batch level, I would prefer to do it in the next PR.

test/Reddit.spec.ts
ChihChengLiang: Just realized since processTx is a view function. The estimate_call don't work. Got "createAccountInstance.processTx.estimate_call is not a function".

ChihChengLiang:
Merging now. Post merging reviews are welcomed

PR #40, opened 2020-07-17
Airdrop to batch transfer 2

ChihChengLiang:


ChihChengLiang:
Move submitBatch and finaliseDepositsAndSubmitBatch back to rollup.sol. As @vaibhavchellani and I discussed offline, the current deployment gas is too costly to have all the functions in one frontend contract. We need both rollup.sol and rollupReddit.sol

PR #44, opened 2020-07-19
fix burn consent check

ChihChengLiang:


contracts/BurnConsent.sol
```diff
@@ -163,13 +163,14 @@ contract BurnConsent is FraudProofHelpers {
             );
         }
 
-        if (!_tx.cancel && account.balance < _tx.amount) {
+        // After the transaction, the user should have enough balance to pay the burn
+        if (!_tx.cancel && account.balance < (_tx.amount + account.burn)) {
```
ChihChengLiang, line 167: @barryWhiteHat is this check the way we want?

ChihChengLiang:
The offline conclusion is that we allow a user to consent any amount of burn. We just don't execute the burn if `account.burn > account.balance`. The check is enforced in BurnExecution.sol.

https://github.com/thehubbleproject/RedditHubble/blob/955ec04d18d5d0f9a48138a2f5be4cdb168dac81/contracts/BurnExecution.sol#L132-L139

PR #45, opened 2020-07-19
add burn execution tests

ChihChengLiang:


contracts/BurnExecution.sol
```diff
@@ -95,51 +93,61 @@ contract BurnExecution is FraudProofHelpers {
...
     }
 
+    function ApplyBurnExecutionTx(
+        Types.AccountMerkleProof memory _fromAccountProof,
+        Types.BurnExecution memory _tx
+    ) public view returns (bytes memory updatedAccount, bytes32 newRoot) {
+        Types.UserAccount memory account = _fromAccountProof.accountIP.account;
+
+        account.balance -= account.burn;
+        account.burn = 0;
```
ChihChengLiang, line 103: @barryWhiteHat do we want to reset burn to 0?
It's a "monthly burn" and the spec does not reset to 0. 
If we don't want to reset to zero, what do we do if we don't have enough balance the next month?

contracts/BurnExecution.sol
```diff
@@ -95,51 +93,61 @@ contract BurnExecution is FraudProofHelpers {
...
      * @notice Overrides processTx in FraudProof
      */
-    function processTx(
+    function processBurnExecutionTx(
         bytes32 _balanceRoot,
-        bytes32 _accountsRoot,
         Types.BurnExecution memory _tx,
-        Types.PDAMerkleProof memory _from_pda_proof,
-        Types.AccountProofs memory accountProofs
+        Types.AccountMerkleProof memory _fromAccountProof
```
ChihChengLiang, line 117: @vaibhavchellani I removed pda proof and the pubkey tree root for now. We might want to add it back depending on how we do the signature validation.

contracts/BurnExecution.sol
barryWhiteHat: So we don't reset to 0. if a user wants to do this they can use authorize burn with amount 0.

If they don't have enough balance the coordinator just does not include them in the batch for that month.

ChihChengLiang: Sounds good. Let me remove the reset and add a balance check.

PR #50, opened 2020-07-20
Fix transfer and add test to it

ChihChengLiang:


contracts/rollupReddit.sol
```diff
@@ -25,6 +25,9 @@ contract RollupReddit {
         airdrop = IReddit(
             nameRegistry.getContractDetails(ParamManager.AIRDROP())
         );
+        transfer = IReddit(
+            nameRegistry.getContractDetails(ParamManager.TRANSFER())
+        );
```
ChihChengLiang, line 30: This is the fix.

PR #55, opened 2020-07-21
BLS Integration

kilic:

PR #63, opened 2020-07-21
Add BLS

ChihChengLiang:
This is a part of #55. We add BLS.sol, AccountTree.sol, BLSAccountRegistry.sol and their tests here.

Sorry to @kilic I lost your committer information while cherry-picking commits.


ChihChengLiang:
Post PR reviews are welcomed

PR #65, opened 2020-07-21
Add Tx.sol

ChihChengLiang:


contracts/libs/Tx.sol
```diff
@@ -0,0 +1,516 @@
...
+    // positions in bytes
+    uint256 public constant POSITION_RECEIVER_3 = 4;
+    uint256 public constant POSITION_AMOUNT_3 = 8;
+
+    struct TransferDecoded {
+        uint256 senderID;
+        uint256 receiverID;
+        uint256 amount;
+        uint256 nonce;
+    }
```
ChihChengLiang, line 60: It looks like our transaction has 3 types of struct: the encoded, the encoded with sig, and the compressed.
Is the TransferDecoded the encoded with sig one that is sent to the fraudproof? Do we need all fields for this? i.e.

```
    struct Transaction {
        uint256 fromIndex;
        uint256 toIndex;
        uint256 tokenType;
        uint256 nonce;
        uint256 txType;
        uint256 amount;
        bytes signature;
    }
```
@vaibhavchellani What's your thoughts on this?

ChihChengLiang:
Merging this now and we fix whatever is required in the next pr

PR #68, opened 2020-07-21
added compress with message for burn txs

vaibhavchellani:


contracts/libs/RollupUtils.sol
```diff
@@ -628,6 +637,15 @@ library RollupUtils {
...
     }
 
+    function CompressBurnExecutionWithMessage(bytes memory message)
+        public
+        pure
+        returns (bytes memory)
+    {
+        Types.BurnExecution memory _tx = BurnExecutionFromBytes(message);
+        return abi.encode(_tx.fromIndex);
+    }
```
ChihChengLiang, line 647: It looks like the input is the same as the output here. But I think it's fine

PR #78, opened 2020-07-23
remove sig from createAccount

ChihChengLiang:


ChihChengLiang:
@vaibhavchellani note we are removing the signature in createAccount as previously discussed.

vaibhavchellani:
Sounds good, thanks!

PR #98, opened 2020-07-25
Add compact burn exec

ChihChengLiang:


ChihChengLiang:
Ewwww ... the `it.only` can skip the whole other testing, scary.

PR #99, opened 2020-07-26
Add Reddit account to genesis leaves

ChihChengLiang:
Approach:

- Hard code Reddit Account at Genesis
- Register Reddit Public key when deploying genesis

### DIFF
--- a/contracts/DepositManager.sol
+++ b/contracts/DepositManager.sol
@@ -63,7 +63,7 @@ contract DepositManager {
         _;
     }
 
-    constructor(address _registryAddr) public {
+    constructor(address _registryAddr, bytes32 redditPubkeyHash) public {
         nameRegistry = Registry(_registryAddr);
         governance = Governance(
             nameRegistry.getContractDetails(ParamManager.Governance())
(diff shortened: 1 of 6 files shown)


contracts/DepositManager.sol
```diff
@@ -79,13 +79,10 @@ contract DepositManager {
...
-        AddCoordinatorLeaves();
-    }
-
-    function AddCoordinatorLeaves() internal {
-        // first 2 leaves belong to coordinator
-        accountsTree.appendLeaf(ZERO_BYTES32);
+        // Coordinator
         accountsTree.appendLeaf(ZERO_BYTES32);
+        // Reddit
+        accountsTree.appendLeaf(redditPubkeyHash);
```
ChihChengLiang, line 85: Register redditPubkeyHash when deploying DepositManager

contracts/libs/RollupUtils.sol
```diff
@@ -113,50 +113,43 @@ library RollupUtils {
...
             burn: 0,
             lastBurn: 0
         });
-        Types.UserAccount memory account2 = Types.UserAccount({
+        accounts[1] = Types.UserAccount({
             ID: 1,
-            tokenType: 0,
-            balance: 0,
+            tokenType: 1,
+            balance: 1000000000000000,
```
ChihChengLiang, line 132: Genesis balance for reddit

contracts/DepositManager.sol
vaibhavchellani: Can we use appendDataBlock here instead, so that it emits an event which the node can catch and register the pubkey?

ChihChengLiang: Sure thing

ChihChengLiang:
We go with the deposit approach now

PR #106, opened 2020-07-27
fix merkle root check

ChihChengLiang:


test/Rollup.spec.ts
```diff
@@ -1072,11 +1072,6 @@ contract("Rollup", async function(accounts) {
             "false ErrorId. it should be `2`"
         );
         await utils.compressAndSubmitBatch(tx, falseResult);
-
-        falseBatchComb.txs.push(txByte);
```
ChihChengLiang, line 1076: This test case is wrong because of this:
1. We attempt to dispute the i-1 batch.
2. We should include the txByte of i-1. 

The deleted line added the txByte of the ith batch to the dispute, thus getting a commitment check failure

PR #110, opened 2020-07-28
fix token type check

ChihChengLiang:
1. We can trust a tokenType in an account is registered because:
  - if the account is created by deposit, we check the tokenType is registered when deposit
  - if the account is created by createAccount, we had a fraud proof for that.

2. We can remove the token check against tx because we can check against from account and to account instead


vaibhavchellani:
Looks good to me.

PR #115, opened 2020-07-29
add compress many encoded bytes utils

ChihChengLiang:
Add utils to compress many encoded Types.<tx_type> to a single compressed bytes.

### DIFF
--- a/contracts/libs/RollupUtils.sol
+++ b/contracts/libs/RollupUtils.sol
@@ -265,6 +265,23 @@ library RollupUtils {
         return Tx.serialize(_txs);
     }
 
+    function CompressManyCreateAccountFromEncoded(bytes[] memory txBytes)
+        public
+        pure
+        returns (bytes memory)
+    {
+        Tx.CreateAccount[] memory _txs = new Tx.CreateAccount[](txBytes.length);
+        for (uint256 i = 0; i < txBytes.length; i++) {
+            Types.CreateAccount memory _tx = CreateAccountFromBytes(txBytes[i]);
+            _txs[i] = Tx.CreateAccount(
+                _tx.accountID,
+                _tx.stateID,
+                _tx.tokenType
+            );
+        }
+        return Tx.serialize(_txs);
+    }
+
     //
     // Airdrop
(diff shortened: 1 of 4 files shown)


contracts/libs/RollupUtils.sol
```diff
@@ -578,6 +613,24 @@ library RollupUtils {
...
+        for (uint256 i = 0; i < txBytes.length; i++) {
+            Types.Transaction memory _tx = TxFromBytes(txBytes[i]);
+            _txs[i] = Tx.Transfer(
+                _tx.fromIndex,
+                _tx.toIndex,
+                _tx.amount,
+                _tx.nonce
+            );
+        }
+        return Tx.serialize(_txs);
```
ChihChengLiang, line 631: To compress many Transfers, we call this function

test/Reddit.spec.ts
```diff
@@ -387,8 +387,8 @@ contract("Reddit", async function() {
         assert.equal(errorCode, ErrorCode.NoError);
         assert.equal(newBalanceRoot, resultTo[1]);
 
-        const compressedTxs = await RollupUtilsInstance.CompressTransferFromEncoded(
-            txBytes
+        const compressedTxs = await RollupUtilsInstance.CompressManyTransferFromEncoded(
+            [txBytes]
```
ChihChengLiang, line 391: An example how to use it

contracts/libs/RollupUtils.sol
vaibhavchellani: So this is for N txs, right?

test/Reddit.spec.ts
vaibhavchellani: Are `txBytes` one transaction?

ChihChengLiang: Yes, it's one transaction

contracts/libs/RollupUtils.sol
ChihChengLiang: Yes, the output is N transactions compressed into a bytes string.

PR #116, opened 2020-07-29
remove nonce check in transfer

ChihChengLiang:
We decided in Monday that we don't commit nonce of the transfer. So no nonce check.

The Raw transfer has nonce field and the user sign the raw transfer with the nonce field.
The compressed transfer has no nonce field, we get nonce from the account in the fraud-proof and the signature should match.

### DIFF
--- a/contracts/FraudProof.sol
+++ b/contracts/FraudProof.sol
@@ -112,10 +112,6 @@ contract FraudProofHelpers is FraudProofSetup {
         uint256 i,
         Types.UserAccount memory _from_account
     ) public pure returns (Types.ErrorCode) {
-        if (txs.transfer_nonceOf(i) != _from_account.nonce.add(1)) {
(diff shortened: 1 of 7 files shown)


contracts/FraudProof.sol
```diff
@@ -107,18 +107,6 @@ contract FraudProofHelpers is FraudProofSetup {
         return Types.ErrorCode.NoError;
     }
 
-    function validateTxBasic(
-        bytes memory txs,
-        uint256 i,
-        Types.UserAccount memory _from_account
-    ) public pure returns (Types.ErrorCode) {
-        if (txs.transfer_nonceOf(i) != _from_account.nonce.add(1)) {
-            return Types.ErrorCode.BadNonce;
```
ChihChengLiang, line 116: We are removing this nonce check

PR #121, opened 2020-07-29
Fix #119, Remove unwanted fields from submit batch

### LINKED ISSUE #119: Remove unwanted fields from submitBatch
We can remove the following:
- [Deposit Root](https://github.com/thehubbleproject/RedditHubble/blob/master/contracts/rollup.sol#L113): We can avoid the storage cost I think, as ZERO_BYTES32 is the default value itself
- [Timestamp](https://github.com/thehubbleproject/RedditHubble/blob/master/contracts/rollup.sol#L118): Has no need
- [StakeCommitted](https://github.com/thehubbleproject/RedditHubble/blob/master/contracts/rollup.sol#L116): I think we can assume that to be constant, we can prob avoid the call to governance contract too.

ChihChengLiang:
- [x] timestamp
- [x] staking amount
- [x] deposit root

### DIFF
--- a/contracts/libs/Types.sol
+++ b/contracts/libs/Types.sol
@@ -31,7 +31,6 @@ library Types {
         bytes32 txRoot;
         uint256 stakeCommitted;
         uint256 finalisesOn;
-        uint256 timestamp;
         Usage batchType;
     }
 
--- a/contracts/libs/Types.sol
+++ b/contracts/libs/Types.sol
@@ -29,8 +29,8 @@ library Types {
         bytes32 depositTree;
         address committer;
         bytes32 txRoot;
-        uint256 stakeCommitted;
         uint256 finalisesOn;
+        bool withdrawn;
         Usage batchType;
     }
 
--- a/contracts/rollup.sol
+++ b/contracts/rollup.sol
@@ -106,16 +106,15 @@ contract RollupHelpers is RollupSetup {
         bytes32 _updatedRoot,
         Types.Usage batchType
     ) internal {
-        Types.Batch memory newBatch = Types.Batch({
-            stateRoot: _updatedRoot,
-            accountRoot: accountsTree.getTreeRoot(),
-            depositTree: bytes32(0x00),
-            committer: msg.sender,
-            txRoot: txRoot,
-            finalisesOn: block.number + governance.TIME_TO_FINALISE(),
-            withdrawn: false,
-            batchType: batchType
-        });
+        Types.Batch memory newBatch;
+        newBatch.stateRoot = _updatedRoot;
+        newBatch.accountRoot = accountsTree.getTreeRoot();
+        // newBatch.depositTree default initialized to 0 bytes
+        newBatch.committer = msg.sender;
+        newBatch.txRoot = txRoot;
+        newBatch.finalisesOn = block.number + governance.TIME_TO_FINALISE();
+        // newBatch.withdrawn default initialized to false
+        newBatch.batchType = batchType;
 
         batches.push(newBatch);
         logger.logNewBatch(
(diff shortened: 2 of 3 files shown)


contracts/libs/Types.sol
```diff
@@ -29,9 +29,8 @@ library Types {
         bytes32 depositTree;
         address committer;
         bytes32 txRoot;
-        uint256 stakeCommitted;
         uint256 finalisesOn;
-        uint256 timestamp;
+        bool withdrawn;
```
ChihChengLiang, line 33: It looks like `stakeCommitted` also acts as the state of the batch being withdrawn or slashed, thus adding a withdrawn field to replace that.
The state of `slashed` is a little bit strange, see next comments

contracts/rollup.sol
```diff
@@ -191,11 +190,9 @@ contract RollupHelpers is RollupSetup {
...
-            uint256 _challengerReward = (batch.stakeCommitted.mul(2)).div(3);
+            uint256 _challengerReward = (STAKE_AMOUNT.mul(2)).div(3);
             challengerRewards += _challengerReward;
-            burnedAmount += batch.stakeCommitted.sub(_challengerReward);
-
-            batches[i].stakeCommitted = 0;
+            burnedAmount += STAKE_AMOUNT.sub(_challengerReward);
 
             // delete batch
             delete batches[i];
```
ChihChengLiang, line 198: This is a little bit strange to me. Why we are setting stakeCommitted = 0 before we are deleting the whole batch?
This is why I consider stakeCommitted does not act as the state of a batch being slashed is not.

vaibhavchellani: yeah, it does seem strange to me too, I am sure there was a reason for that which I should have commented, but for now, let's consider it a mistake.

contracts/libs/Types.sol
```diff
@@ -26,12 +26,10 @@ library Types {
     struct Batch {
         bytes32 stateRoot;
         bytes32 accountRoot;
-        bytes32 depositTree;
```
ChihChengLiang, line 29: I'm super uncertain about this. Haven't given it a good thought

vaibhavchellani: Yeah we dont want to remove that, we need it when we do rollback, all we want to do is remove [this](https://github.com/thehubbleproject/RedditHubble/blob/master/contracts/rollup.sol#L113) line as it will pick up the default zero value automatically

ChihChengLiang: Ah ah I see.

contracts/rollup.sol
```diff
@@ -107,17 +106,15 @@ contract RollupHelpers is RollupSetup {
...
-        });
+        Types.Batch memory newBatch;
+        newBatch.stateRoot = _updatedRoot;
+        newBatch.accountRoot = accountsTree.getTreeRoot();
+        // newBatch.depositTree default initialized to 0 bytes
+        newBatch.committer = msg.sender;
+        newBatch.txRoot = txRoot;
+        newBatch.finalisesOn = block.number + governance.TIME_TO_FINALISE();
+        // newBatch.withdrawn default initialized to false
+        newBatch.batchType = batchType;
```
ChihChengLiang, line 117: I compared these two different styles of assignment and the incoming one is slightly more gas efficient.


```solidity
pragma solidity ^0.5.15;


contract Foo {
    struct A {
        bytes32 a;
        bytes32 b;
        bytes32 c;
    }
    // 478 gas
    function createStruct () public pure{
        A memory a;
        a.a = keccak256("123");
        a.b = keccak256("456");
  
    }
    // 535 gas
    function createStruct2() public pure {
        A memory a = A({
            a: keccak256("123"),
            b: keccak256("456"),
            c: bytes32(0x00)
        });
    }
}
```

ChihChengLiang:
The submitbatch now costs 188K gas, as the number in [master branch](https://github.com/thehubbleproject/RedditHubble/runs/924249053?check_suite_focus=true) is 230K

contracts/rollup.sol
vaibhavchellani: Perfect!

PR #122, opened 2020-07-30
add decompress functions

ChihChengLiang:


contracts/libs/RollupUtils.sol
```diff
@@ -282,6 +282,19 @@ library RollupUtils {
         return Tx.serialize(_txs);
     }
 
+    function DecompressManyCreateAccount(bytes memory txs)
+        public
+        pure
+        returns (Tx.CreateAccount[] memory structTxs)
```
ChihChengLiang, line 288: Not sure if decompress many works for the go client

contracts/libs/RollupUtils.sol
```diff
@@ -437,6 +450,14 @@ library RollupUtils {
         return Tx.serialize(_txs);
     }
 
+    function DecompressManyAirdrop(bytes memory txs)
+        public
+        pure
+        returns (Tx.Transfer[] memory structTxs)
+    {
+        return DecompressManyTransfer(txs);
```
ChihChengLiang, line 458: Reuse Transfer's for airdrop, because we use Tx.Transfer for them for the time being.

vaibhavchellani:
Looks great to me!

PR #123, opened 2020-07-30
Transaction messages to sign

kilic:


contracts/FraudProof.sol
```diff
@@ -198,31 +198,32 @@ contract FraudProofHelpers is FraudProofSetup {
...
-                    _tx
-                        .nonce,
-                    _tx
-                        .amount
-                )
-                    .ecrecovery(_tx.signature),
-            "Signature is incorrect"
-        );
-    }
+    // TODO: do we need this?
```
ChihChengLiang, line 201: Remove commented

contracts/libs/RollupUtils.sol
```diff
@@ -812,14 +815,6 @@ library RollupUtils {
         return _tx;
     }
 
-    function BurnExecutionSignBytes(uint256 txType, uint256 fromIndex)
```
ChihChengLiang, line 815: Need to restore this for the time being for the tests to pass

contracts/libs/RollupUtils.sol
```diff
@@ -690,14 +687,20 @@ library RollupUtils {
...
-        uint256 amount,
-        uint256 nonce
+        uint256 toIndex,
+        uint256 nonce,
+        uint256 amount
     ) public pure returns (bytes32) {
         return
             keccak256(
-                BytesFromBurnConsentNoStruct(txType, fromIndex, amount, nonce)
+                abi.encode(
```
vaibhavchellani, line 697: Why call abi.encode instead of reusing the util?

contracts/libs/RollupUtils.sol
```diff
@@ -531,20 +531,17 @@ library RollupUtils {
     }
 
     function getTxSignBytes(
-        uint256 txType,
```
vaibhavchellani, line 534: Can we change this to TransferSignBytes or something like that?

contracts/libs/RollupUtils.sol
```diff
@@ -692,12 +681,12 @@ library RollupUtils {
     function BurnConsentSignBytes(
         uint256 txType,
         uint256 fromIndex,
-        uint256 amount,
-        uint256 nonce
+        uint256 nonce,
+        uint256 amount
```
ChihChengLiang, line 685: @vaibhavchellani want to confirm if we want this

vaibhavchellani: Should be okay, ill update the client. But thanks for pointing out.

PR #129, opened 2020-07-31
add sig verification for transfer

ChihChengLiang:


contracts/libs/Tx.sol
```diff
@@ -791,6 +791,43 @@ library Tx {
...
+    function airdrop_verify(
+        bytes memory txs,
+        uint256 index,
+        uint256 nonce,
+        address signer
+    ) internal pure returns (bool) {
+        Transfer memory _tx = transfer_decode(txs, index);
+        bytes32 message = keccak256(
+            abi.encodePacked(
+                AIRDROP,
```
ChihChengLiang, line 821: Since we hardcode Txtype here, we can't use transfer_verify directly in airdrop.sol. That's why we are adding airdrop_verify.

PR #130, opened 2020-07-31
Reduces storage usage in Batch

ChihChengLiang:


contracts/rollup.sol
```diff
@@ -86,40 +86,42 @@ contract RollupSetup {
...
+        );
         newBatch.committer = msg.sender;
-        newBatch.txRoot = txRoot;
         newBatch.finalisesOn = block.number + governance.TIME_TO_FINALISE();
         // newBatch.withdrawn default initialized to false
         newBatch.batchType = batchType;
 
         batches.push(newBatch);
-        logger.logNewBatch(
+        emit logger.NewBatch(
```
vaibhavchellani, line 120: You prob will need to remove `emit` here since you are calling another function.

PR #133, opened 2020-07-31
Store intermediate subtree node only once

kilic:


ChihChengLiang:
I think I finally got how this works. Great improvement! We are already using it in the current AccountTree.

Below is my understanding of how the intermediate subtree works. 

---

Let's say we are inserting a leaf path `1001` to the tree. * are leaves already in the tree, @ is the leaf to insert, and . is empty


i=0 the 0th bit is odd, we hash with left to get acc
- the left is the previous filled subtree at 0
```
                acc
                / \
* * * * * * * * * @ . . . . . .
```
i=1 the 1st bit is even, we hash with right to get acc'.
- The right is a zero[1]. 
- And we set the `filledSubtreesLeft[1]` to acc
```
                   acc'
                  / \
                acc zero[1]
                / \
* * * * * * * * * @ . . . . . .
```

i=2 the 2nd bit is even, we hash with right.

- Should we update `filledSubtreesLeft[2]` to acc'? (spoiler: no)

```
                      acc''
                    /    \
                   acc'  zero[2]
                  / \
                acc zero[1]
                / \
* * * * * * * * * @ . . . . . .
```

Watch how we insert the next leaf `1010`

i = 0 0th bit is even, we hash with right to get acc2.

```
                   acc'
                  / \
                acc acc2
                / \  /
* * * * * * * * * @ @' . . . . .
```

i = 1 1st bit is odd, we hash with left to get acc2'.
- Here we see the previous `filledSubtreesLeft[1]` (acc) is reused, but not the `filledSubtreesLeft[2]` (acc') because it is replaced by acc2'. 

```
                   acc2'
                  / \
                acc acc2
                / \  /
* * * * * * * * * @ @' . . . . .
```

PR #142, opened 2020-08-11
Migrate to buidler

ChihChengLiang:
I like the deploying speed of buidler.


ChihChengLiang:
The testing time is reduced from 5 min to 53 sec.

test/Deposit.spec.ts
```diff
@@ -12,7 +13,7 @@ const RollupUtils = artifacts.require("RollupUtils");
 import { StateStore } from "../scripts/helpers/store";
 chai.use(chaiAsPromised);
 
-contract("DepositManager", async function(accounts) {
+contract("DepositManager", async function() {
```
ChihChengLiang, line 16: If you need to use `Contract.deployed()` use contract. It runs `truffle-fixture.js`

test/account_tree.test.ts
```diff
@@ -4,7 +4,7 @@ import { Tree, Hasher } from "./utils/tree";
 
 let DEPTH: number;
 let BATCH_DEPTH: number;
-contract("Account Tree", accounts => {
+describe("Account Tree", async () => {
```
ChihChengLiang, line 7: If you don't use `Contract.deployed()` but use `Contract.new()`, use `describe`. It doesn't run truffle-fixture.js

test/truffle-fixture.js
```diff
@@ -0,0 +1,198 @@
+// Libs
```
ChihChengLiang, line 1: truffle-fixture.js mirrors the `./migrations/2_deploy.js`. I still keep the `./migrations/2_deploy.js` in case we need to deploy something. I'll move to custom deploy functions and completely remove `./migrations/` this week so that we have the same deploy process in testing and in production.

PR #146, opened 2020-08-12
BLS Integration 2nd attempt

kilic:


contracts/DepositManager.sol
```diff
@@ -75,53 +73,21 @@ contract DepositManager {
...
 
     /**
      * @notice Adds a deposit for an address to the deposit queue
-     * @param _destination Address for which we are depositing
      * @param _amount Number of tokens that user wants to deposit
      * @param _tokenType Type of token user is depositing
      */
     function depositFor(
-        address _destination,
+        uint256 accountID,
```
vaibhavchellani, line 84: How do I get an account ID to send here?

vaibhavchellani:
LGTM.

contracts/DepositManager.sol
kilic: Depositor has to be registered to BLSAccountRegistery in advance

kilic:
> LGTM. Need to fix tests.

PR is not ready to merge

PR #153, opened 2020-08-15
use utils on transfer

ChihChengLiang:


contracts/libs/Tx.sol
```diff
@@ -431,15 +431,13 @@ library Tx {
         for (uint256 i = 0; i < txs.length; i++) {
             uint256 fromIndex;
             uint256 amount;
-            bytes memory signature;
-            (, fromIndex, amount, , signature) = abi.decode(
+            (, fromIndex, amount, ) = abi.decode(
                 txs[i],
-                (uint256, uint256, uint256, uint256, bytes)
```
ChihChengLiang, line 437: Remove the signatures of the burn consent

test/registry.test.ts
```diff
@@ -12,11 +13,9 @@ type Pubkey = mcl.mclG2;
...
 function pubkeyToLeaf(p: Pubkey) {
     const uncompressed = mcl.g2ToHex(p);
-    const leaf = web3.utils.soliditySha3(
-        { t: "uint256", v: uncompressed[0] },
-        { t: "uint256", v: uncompressed[1] },
-        { t: "uint256", v: uncompressed[2] },
-        { t: "uint256", v: uncompressed[3] }
+    const leaf = ethers.utils.solidityKeccak256(
+        ["uint256", "uint256", "uint256", "uint256"],
+        uncompressed
```
ChihChengLiang, line 18: The typescript complains `soliditySha3` might return null type, so I changed those web3.utils.soliditySha3 to ethers.utils.solidityKeccak256.

test/utils/tx.ts
```diff
@@ -15,6 +14,9 @@ function log2(n: number) {
 export interface Tx {
     hash(): string;
     encode(prefix?: boolean): string;
+}
+
+export interface SignableTx extends Tx {
     message(domain: string): string;
 }
```
ChihChengLiang, line 21: Differentiate Tx and SignableTx.

PR #162, opened 2020-08-16
Fix rollup test

ChihChengLiang:


scripts/deploy.ts
```diff
@@ -0,0 +1,14 @@
...
+async function main() {
+    const provider = new ethers.providers.JsonRpcProvider();
+    const signer = provider.getSigner();
+
+    const allContracts = await deployAll(signer, TESTING_PARAMS);
+    Object.keys(allContracts).forEach((contract: string) => {
+        console.log(contract, allContracts[contract].address);
+    });
+}
+main();
```
ChihChengLiang, line 14: Move this main function to scripts. Since the main() breaks the tests

test/Rollup.spec.ts
```diff
@@ -1,377 +1,163 @@
...
+        const depth = 1; // Math.log2(commitmentLength + 1)
+        const tree = Tree.new(
+            depth,
+            Hasher.new(
+                "bytes",
+                ethers.utils.keccak256(
+                    "0x0000000000000000000000000000000000000000000000000000000000000000"
+                )
+            )
+        );
```
ChihChengLiang, line 94: Here highlighted some behaviors of `MerkleTreeUtils.getMerkleRootFromLeaves`, which we used in rollup contract to get Merkle root of commitments. Might be gotchas for clients.

- It uses `hash(ZERO_BYTES)` as zero leaf.
- It uses minimum possible height, depending on the length of the leaves.

test/utils/account_tree.ts
```diff
@@ -59,3 +60,61 @@ export class AccountRegistry {
         return leaf;
     }
 }
+
+export class AccountRegistry2 {
```
ChihChengLiang, line 64: Creating this class to sync with Ethers instance

PR #163, opened 2020-08-17
add deploy script

ChihChengLiang:


vaibhavchellani:
To summarise, this PR adds scripts that register and does a deposit for 2 accounts along with initializing the contract with a genesis state root. Is the summary correct?

ChihChengLiang:
It registers 2 public keys and initializes a genesis state root. But no deposit.

PR #167, opened 2020-08-20
bring back constructors

ChihChengLiang:


contracts/Transfer.sol
```diff
@@ -142,24 +142,24 @@ contract Transfer is FraudProofHelpers {
...
+     * Constructor *
+     ********************/
+    constructor(address _registryAddr) public {
+        nameRegistry = Registry(_registryAddr);
+
+        governance = Governance(
+            nameRegistry.getContractDetails(ParamManager.Governance())
+        );
+
+        merkleUtils = MTUtils(
```
ChihChengLiang, line 155: If you can't pass the fraudproof test, one of the reasons might be the merkleUtils here is disabled before.

ChihChengLiang:
Almost outdated

PR #170, opened 2020-08-21
fix bls profiling

ChihChengLiang:


ChihChengLiang:
The changing of test contract functions to "view" functions is not desired. Working on a emit event approach.

PR #173, opened 2020-08-23
Fix constructors

ChihChengLiang:
### What's wrong

We have the weird nameRegistry thing. So every time we deploy a fraudproof contract, we have to do this:
```
deploy nameRegistry
deploy ParamManager
deploy governance
deploy merkleUtils
deploy tokenRegistry
await nameRegistry.register(ParamManager.governance, governance.address)
await nameRegistry.register(ParamManager.merkleUtils, merkleUtils.address)
await nameRegistry.register(ParamManager.tokenRegistry, tokenRegistry.address)
```
That's very annoying and bad for unit tests.

### How it is fixed

Move the existing constructors to inherit contracts in contracts/prod. We can use much simpler constructor in tests.

```
// ./contracts/prod
contract TransferProduction is Transfer {
    constructor(_registry){
        // complicate setup with nameRegistry and ParamManager
    }
}
```

```
// ./contracts/test
contract TestTransfer is Transfer {
    constructor(merkleUtils, governance){
        // simple setup
    }
}
```


vaibhavchellani:
Umm..not sure if this is a good standard to follow, do any projects do this right now?

ChihChengLiang:
per offline discussion, we accept this as a mid-term solution.

PR #176, opened 2020-08-24
remove pdf proof from process batch

ChihChengLiang:
Since we are checking the public key when disputing signature


kilic:
Yay!

Also removed the idle struct.

PR #184, opened 2020-08-26
Go full ethers

ChihChengLiang:
and waffle too


buidler.config.ts
```diff
@@ -2,8 +2,7 @@ import { usePlugin } from "@nomiclabs/buidler/config";
 import { generateFirstWallets, mnemonics } from "./scripts/helpers/wallet";
 import { ethers } from "ethers";
 
-usePlugin("@nomiclabs/buidler-truffle5");
-usePlugin("@nomiclabs/buidler-ethers");
+usePlugin("@nomiclabs/buidler-waffle");
```
ChihChengLiang, line 5: Tried waffle and not breaking anything, so I leave it there.

contracts/test/TestTransfer.sol
```diff
@@ -6,26 +6,29 @@ import { Types } from "../libs/Types.sol";
...
+            signature,
+            proof,
+            stateRoot,
+            accountRoot,
+            appID,
+            txs
         );
+
+        emit Return1(operationCost - gasleft());
+        emit Return2(err);
```
ChihChengLiang, line 31: Here's how we get gas costs and result together. The `Types.ErrorCode err` assignment might bias the result a little bit but should be bearable. cc @kilic

test/account_tree.test.ts
```diff
@@ -37,7 +41,7 @@ describe("Account Tree", async () => {
     });
     it("update with single leaf", async function() {
         for (let i = 0; i < 33; i++) {
-            const leaf = web3.utils.randomHex(32);
+            const leaf = ethers.utils.hexlify(ethers.utils.randomBytes(32));
```
ChihChengLiang, line 44: `ethers.utils.randomBytes` returns Uint8Array for bytes manipulation. But the current ts utils are written to support strings only. And we are doing all the weird `slice` and `"0x" +` to manipulate hex strings.
I'll push more commits to fix this.

ChihChengLiang: left the task here https://github.com/thehubbleproject/RedditHubble/issues/186

PR #188, opened 2020-09-01
added aggregated sig to commitment

### LINKED ISSUE #157: Aggregated signature not part of commitment yet
We haven't included aggregated signature as part of the commitment yet, we just accept it as an argument at the moment, need to commit it as well.

vaibhavchellani:
Fixes #157

### DIFF
--- a/contracts/libs/RollupUtils.sol
+++ b/contracts/libs/RollupUtils.sol
@@ -18,7 +18,13 @@ library RollupUtils {
     ) public pure returns (bytes32) {
         return
             keccak256(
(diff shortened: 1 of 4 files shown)


ChihChengLiang:
Why don't we add a field in commitment struct? The change  in the current approach looks smaller though, but also a little bit implicit.

vaibhavchellani:
Yeah can do that too. Just would need to propagate one more variable to [dispute](https://github.com/thehubbleproject/RedditHubble/blob/master/contracts/rollup.sol#L320) I think.

ChihChengLiang (approved):
Looks good to me

PR #189, opened 2020-09-01
Fix rollup integration test

ChihChengLiang:


contracts/rollup.sol
```diff
@@ -243,7 +243,7 @@ contract Rollup is RollupHelpers {
                 RollupUtils.CommitmentToHash(
                     updatedRoots[i],
                     pubkeyTreeRoot,
-                    keccak256(abi.encode(txs[i])),
+                    keccak256(txs[i]),
```
ChihChengLiang, line 246: This diff matters, it affects test result

test/Rollup.spec.ts
```diff
@@ -65,19 +56,18 @@ describe("Rollup", async function() {
 
         const signature = Alice.sign(tx);
 
-        const rollup = contracts.rollup as Rollup;
-        const rollupUtils = contracts.rollupUtils as RollupUtils;
+        const rollup = contracts.rollup;
+        const rollupUtils = contracts.rollupUtils;
```
ChihChengLiang, line 60: These were the wrong assignment of types, and they matter.

PR #190, opened 2020-09-01
Rm unused type

ChihChengLiang:
- Replace Types.Transaction with Types.Transfer
- The old compress API is unused now


ChihChengLiang:
Got an offline approval from @vaibhavchellani

PR #192, opened 2020-09-01
Add fee calculations

ChihChengLiang:


contracts/Transfer.sol
```diff
@@ -118,21 +119,32 @@ contract Transfer is FraudProofHelpers {
...
             if (!isTxValid) {
                 break;
             }
         }
+        if (isTxValid) {
+            (stateRoot, , isTxValid) = processFee(
+                stateRoot,
+                fees,
+                feeReceiver,
+                accountProofs[length * 2]
```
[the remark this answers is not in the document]
vaibhavchellani: I think this is the way to go for now.

contracts/Transfer.sol
```diff
@@ -95,26 +95,41 @@ contract Transfer is FraudProofHelpers {
     function processTransferBatch(
         bytes32 stateRoot,
         bytes memory txs,
-        Types.AccountMerkleProof[] memory accountProofs
+        Types.AccountMerkleProof[] memory accountProofs,
+        uint256 tokenType,
+        uint256 feeReceiver
```
ChihChengLiang, line 100: We require the transactions in this commitment to have the same tokenType, so that we can guarantee the fee receiver receives the same types of the token.

contracts/libs/Types.sol
```diff
@@ -33,11 +33,20 @@ library Types {
         bool withdrawn;
     }
 
+    struct Submission {
+        bytes txs;
+        bytes32 updatedRoot;
+        uint256[2] signature;
+        uint256 tokenType;
+    }
```
ChihChengLiang, line 41: Add Submission struct to prevent stack too deep in the submitBatch.

contracts/rollup.sol
```diff
@@ -230,21 +234,22 @@ contract Rollup is RollupHelpers {
...
-        uint256 commmitmentLength = updatedRoots.length;
-        bytes32[] memory commitments = new bytes32[](commmitmentLength);
-        for (uint256 i = 0; i < commmitmentLength; i++) {
+        bytes32[] memory commitments = new bytes32[](submissions.length);
+        bytes32 pubkeyTreeRoot = accountRegistry.root();
+        for (uint256 i = 0; i < submissions.length; i++) {
             commitments[i] = (
                 RollupUtils.CommitmentToHash(
-                    updatedRoots[i],
-                    accountRegistry.root(),
```
ChihChengLiang, line 245: Before this PR the `accountRegistry.root()` was called n times.

contracts/Transfer.sol
```diff
@@ -95,26 +95,41 @@ contract Transfer is FraudProofHelpers {
...
         for (uint256 i = 0; i < length; i++) {
             // call process tx update for every transaction to check if any
             // tx evaluates correctly
+            _tx = txs.transfer_decode(i);
+            fees = fees.add(_tx.fee);
             (stateRoot, , , , isTxValid) = processTx(
                 stateRoot,
-                txs.transfer_decode(i),
+                _tx,
+                tokenType,
```
ChihChengLiang, line 116: We get this tokenType all the way from the commitment to validate each transaction sender's state leaf has the same token type as the fee receiver's. We can also get the tokenType from the fee receiver's state leaf merkle proof. What do you think?

contracts/libs/Types.sol
vaibhavchellani: Can we just use commitments struct?

contracts/Transfer.sol
vaibhavchellani: I think it's better to validate the fee receiver's state leaf and then cache the token type, post that we can check the token type of senders against this cached fee receiver's token type. I believe that's what is happening now?

ChihChengLiang: Not yet. Let me do this

contracts/libs/Types.sol
ChihChengLiang: I considered commitments struct too, but the commitment has extra fields that are set inside the function body of submitBatch. Using the commitment struct:
1. waste extra zero fields on the input.
2. confuses the developer because you send input that has no effect.

vaibhavchellani: Gotcha, cool let's roll with this.

contracts/Transfer.sol
ChihChengLiang: After some trying, I realized this would come with the cost of adding 1 witness in the fraudproof. Detail reasoning in #206. Will stick to committing the tokenType now and defer the decision later.

PR #193, opened 2020-09-01
Flatten merkle data

ChihChengLiang:


contracts/libs/Types.sol
```diff
@@ -106,7 +93,8 @@ library Types {
     }
 
     struct AccountMerkleProof {
-        AccountInclusionProof accountIP;
+        UserAccount account;
+        uint256 pathToAccount; // This field is kept for backward competibility, don't use it.
```
ChihChengLiang, line 97: We keep the pathToAccount for backward compatibility. Basically this value can't be trusted as it can be used to create ghost proof.

vaibhavchellani: Feel free to remove it, I will update the client accordingly

ChihChengLiang: I didn't remove it because many Reddit functions are depending on it, and it takes some work to fix them all. It might not be worth it to fix them now.

vaibhavchellani (approved):
LGTM

PR #196, opened 2020-09-03
Attempt to upgrade to ethers v5

ChihChengLiang:
We want ethers v5 for the callstatic function. Which is useful in profiling gas.

Blocker is the typechain-etherv5. Typescript complains the generated factory doesn't work for the base class ContractFactory

```
Property 'deploy' in type 'AirdropFactory' is not assignable to the same property in base type 'ContractFactory'.
```

need to rewrite the deploy script.


ChihChengLiang:
#197 works better

PR #197, opened 2020-09-03
Upgrade ethers v5 take 2

ChihChengLiang:


contracts/rollup.sol
```diff
@@ -45,7 +44,6 @@ interface IRollupReddit {
 
 contract RollupSetup {
     using SafeMath for uint256;
-    using BytesLib for bytes;
```
ChihChengLiang, line 48: It looks like we didn't use any of BytesLib's API

Issue #177, opened 2020-08-24
Incorrect state root comparison after transition


kilic:
After transition is performed with committed transactions comparison is applyied against __old__ stateRoot.
 https://github.com/thehubbleproject/RedditHubble/blob/59affd3e26ac8c2f9c9ff945ad3b376e9e3d1c68/contracts/rollup.sol#L367
https://github.com/thehubbleproject/RedditHubble/blob/59affd3e26ac8c2f9c9ff945ad3b376e9e3d1c68/contracts/rollup.sol#L387

How it should be like:

* Case 1, dispute targeted 0th commitment: we need to compare against final state root of previous batch.
* Case 2, dispute targeted nth commitment: then we must compare against (n-1)th state root of same batch.

With both cases we need to reveal two consecutive state roots commitements in order to detect invalid transition

vaibhavchellani:
Ah! Yep, this is a valid bug.

PR #199, opened 2020-09-04
Mass Migrations

vaibhavchellani:
This PR implements mass migrations for hubble using the spec: https://hackmd.io/pZvHalDsTeah8acNi3fveg

There are 3 things missing in this PR:
- WithdrawRoot check
- The signature check-in claim token
- Test case for the receiver spoke


contracts/MassMigs.sol
```diff
@@ -0,0 +1,170 @@
+pragma solidity ^0.5.15;
+pragma experimental ABIEncoderV2;
+
+import { FraudProofHelpers } from "./FraudProof.sol";
+import { Types } from "./libs/Types.sol";
+import { RollupUtils } from "./libs/RollupUtils.sol";
+import { Tx } from "./libs/Tx.sol";
+import { MerkleTreeUtilsLib } from "./MerkleTreeUtils.sol";
+import "@nomiclabs/buidler/console.sol";
```
ChihChengLiang, line 9: should remove console.log
```suggestion
```

contracts/MassMigs.sol
```diff
@@ -0,0 +1,170 @@
...
+
+contract MassMigs is FraudProofHelpers {
```
ChihChengLiang, line 11: Can we rename all "MassMigs" to "MassMigration"? Want the variables to be easy to pronounce so we can communicate better.

These places need the rename:

- MassMigs.sol
- contract MassMigs
- MassMig struct in Tx.sol

contracts/libs/Tx.sol
```diff
@@ -809,4 +828,33 @@ library Tx {
                 abi.encodePacked(BURN_CONCENT, _tx.fromIndex, nonce, _tx.amount)
             );
     }
+
+    function mass_migration_decode(bytes memory txs, uint256 index)
```
ChihChengLiang, line 832: The convention here should be massMigration_decode
```suggestion
    function massMigration_decode(bytes memory txs, uint256 index)
```

contracts/libs/Tx.sol
```diff
@@ -809,4 +828,33 @@ library Tx {
...
+                MASK_STATE_ID
+            )
+            amount := and(mload(add(p_tx, POSITION_AMOUNT_5)), MASK_AMOUNT)
+            spokeID := and(mload(add(p_tx, POSITION_SPOKE_5)), MASK_SPOKE)
+            fee := and(mload(add(p_tx, POSITION_FEE_0)), MASK_FEE)
+        }
+        return MassMig(sender, receiver, amount, spokeID, fee);
+    }
+
+    function mass_mig_size(bytes memory txs) internal pure returns (uint256) {
```
ChihChengLiang, line 857: The convention here should be massMigration_size

```suggestion
    function massMigration_size(bytes memory txs) internal pure returns (uint256) {
```

test/MassMigs.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+    });
+
+    it("submit a batch and dispute", async function() {
+        const tx = new TxMassMig(Alice.stateID, 0, 5, 1, 1, Alice.nonce + 1);
+        const signature = Alice.sign(tx);
+        const rollup = contracts.rollup;
+        const rollupUtils = contracts.rollupUtils;
+        var stateRoot = stateTree.root;
+        const proof = stateTree.applyMassMigration(tx);
+        // stateRoot = stateTree.root;
```
ChihChengLiang, line 58:
```suggestion
```

test/MassMigs.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+        stateTree.createAccount(Alice);
+        stateTree.createAccount(Bob);
+    });
+
+    it("submit a batch and dispute", async function() {
+        const tx = new TxMassMig(Alice.stateID, 0, 5, 1, 1, Alice.nonce + 1);
+        const signature = Alice.sign(tx);
+        const rollup = contracts.rollup;
+        const rollupUtils = contracts.rollupUtils;
+        var stateRoot = stateTree.root;
```
ChihChengLiang, line 56: We should use `const` whenever we can and use `let` whenever the variable would be reassigned.
We should never use `var`

```suggestion
        const stateRoot = stateTree.root;
```

test/MassMigs.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+        const proof = stateTree.applyMassMigration(tx);
+        // stateRoot = stateTree.root;
+        const txs = ethers.utils.arrayify(tx.encode(true));
+        console.log("transaction", txs);
```
ChihChengLiang, line 60:
```suggestion
```

test/MassMigs.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+
+        const commitment = {
+            stateRoot,
+            accountRoot: root,
+            txHashCommitment: ethers.utils.solidityKeccak256(["bytes"], [txs]),
+            massMigrationMetaInfo: MMInfo,
+            signature: aggregatedSignature0,
+            batchType: Usage.MassMigration
+        };
+        var result = await contracts.rollupReddit.processMMBatch(
```
ChihChengLiang, line 79:
```suggestion
        const result = await contracts.rollupReddit.processMMBatch(
```

ts/stateTree.ts
```diff
@@ -55,6 +55,12 @@ interface ProofCreateAccountTx {
 }
 type ProofCreateAccountBatch = ProofCreateAccountTx[];
 
+interface ProofOfMassMigTx {
```
ChihChengLiang, line 58:
```suggestion
interface ProofOfMassMigrationTx {
```

ts/tx.ts
```diff
@@ -211,3 +213,70 @@ export class TxBurnExecution implements Tx {
         return encoded;
     }
 }
+
+export class TxMassMig implements SignableTx {
```
ChihChengLiang, line 217:
```suggestion
export class TxMassMigration implements SignableTx {
```

contracts/libs/Tx.sol
```diff
@@ -809,4 +828,33 @@ library Tx {
                 abi.encodePacked(BURN_CONCENT, _tx.fromIndex, nonce, _tx.amount)
             );
     }
+
```
ChihChengLiang, line 831: Should add tests in `txSerialization.test.ts` for `massMigration_decode` and `massMigration_size`.

test/MassMigs.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+import * as mcl from "../ts/mcl";
+import { Tree, Hasher } from "../ts/tree";
+import { allContracts } from "../ts/allContractsInterfaces";
+import { assert } from "chai";
+import { parseEvents } from "../ts/utils";
+
+const DOMAIN =
+    "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff";
+
+describe("Rollup", async function() {
```
ChihChengLiang, line 18:
```suggestion
describe("MassMigration", async function() {
```

test/MassMigs.spec.ts
```diff
@@ -0,0 +1,174 @@
+import { Usage } from "../ts/interfaces";
```
ChihChengLiang, line 1: Can we name the file `MassMigration.test.ts`?

ChihChengLiang (commented):
Great work!
Here's my first pass of the review. Left some comments for style and naming stuff.

vaibhavchellani:
Thanks for the review and comments @ChihChengLiang, will make these changes and push asap!

contracts/libs/Types.sol
```diff
@@ -51,12 +52,34 @@ library Types {
         Usage batchType;
     }
 
+    struct MMCommitment {
+        bytes32 stateRoot;
+        bytes32 accountRoot;
+        bytes32 txHashCommitment;
```
ChihChengLiang, line 58: If we can use txs here instead of txHashCommitment, we can save some other inputs and checks. Since we committed the whole MMCommitment, no point to hash txs here anymore.

contracts/Vault.sol
```diff
@@ -0,0 +1,90 @@
...
+
+    constructor(address _registryAddr) public {
+        nameRegistry = Registry(_registryAddr);
+        rollup = Rollup(
+            nameRegistry.getContractDetails(ParamManager.ROLLUP_CORE())
+        );
+        tokenRegistry = ITokenRegistry(
+            nameRegistry.getContractDetails(ParamManager.TOKEN_REGISTRY())
+        );
+        merkleUtils = MTUtils(
```
ChihChengLiang, line 28: We don't have to use this on chain merkleUtils. We only use `verifyLeaf` here, so use MerkleTreeUtilsLib library is cheaper.
```
import { MerkleTreeUtilsLib } from "./MerkleTreeUtils.sol";
```

contracts/Rollup.sol
```diff
@@ -403,6 +457,86 @@ contract Rollup is RollupHelpers {
...
+                commitmentMP.commitment.txHashCommitment != ZERO_BYTES32,
+                "Cannot dispute blocks with no transaction"
+            );
+        }
+
+        bytes32 updatedBalanceRoot;
+        bool isDisputeValid;
+        bytes32 txRoot;
+        (updatedBalanceRoot, txRoot, isDisputeValid) = rollupReddit
+            .processMMBatch(commitmentMP.commitment, txs, accountProofs);
```
ChihChengLiang, line 520: [later] I think this has the same problem as #177. The state root in the commitment is the "post" state root, while the processBatch needs "pre" state root.

contracts/Rollup.sol
```diff
@@ -269,6 +282,47 @@ contract Rollup is RollupHelpers {
...
+                    MMInfo[i].amount,
+                    MMInfo[i].withdrawRoot,
+                    MMInfo[i].targetSpokeID,
+                    aggregatedSignatures[i],
+                    Types.Usage.MassMigration
+                )
+            );
+        }
+        Types.Batch memory newBatch = Types.Batch({
+            commitmentRoot: merkleUtils.getMerkleRootFromLeaves(commitments),
```
ChihChengLiang, line 311: [later] I think we'll have a problem in normal disputeBatch too. The normal disputeBatch should look for previous commitment for the pre state root. If the previous commitment happens to be a mass migration commitment, we can't pass the mass migration commitment to the current disputeBatch since the function arguments have different types. Need to fix this somehow.

contracts/MassMigrations.sol
```diff
@@ -0,0 +1,169 @@
...
+            newRoot,
+            new_from_account,
+            new_to_account,
+            fromAccountProof.account.tokenType,
+            Types.ErrorCode.NoError,
+            true
+        );
+    }
+
+    function ApplyMassMigTxSender(
```
ChihChengLiang, line 154: Nitpick
```suggestion
    function ApplyMassMigrationTxSender(
```

ChihChengLiang (approved):
Left some nitpicks.
Other than those it looks good to me.
I also find some bigger issues that look too involving to fix in this PR. Maybe it's better to address those later. Those issues are marked with `[later]`.

contracts/libs/Types.sol
vaibhavchellani: Done!

contracts/MassMigrations.sol
vaibhavchellani: Done

contracts/Vault.sol
vaibhavchellani: Let me fix this in the next PR.

PR #201, opened 2020-09-06
Fouque Tibouchi Hash to Curve

kilic:
This pr is an attemption to 

- [x] remove unused bls functions
- [x] migrate to Fouque Tibouchi hash to curve
- [x] use [faster modexp](https://github.com/ChihChengLiang/modexp) functions for sqrt and inversion


kilic:
Signature check for 32 tx.
operation gas cost: 5072117
transaction gas cost: 6820242

contracts/libs/BLS.sol
```diff
@@ -95,34 +105,94 @@ library BLS {
...
+        bool success;
+        // solium-disable-next-line security/no-inline-assembly
+        assembly {
+            success := staticcall(sub(gas(), 2000), 6, bnAddInput, 128, p0, 64)
+            switch success
+                case 0 {
+                    invalid()
+                }
+        }
+        require(success, "");
```
ChihChengLiang, line 130: Should give reason here for better debugging.
```suggestion
        require(success, "Hash to point fail");
```

contracts/libs/BLS.sol
```diff
@@ -95,34 +105,94 @@ library BLS {
...
-            y = mulmod(y, x, N);
-            y = addmod(y, 3, N);
-            (y, found) = sqrt(y);
-            if (found) {
-                p[0] = x;
-                p[1] = y;
-                break;
+        require(_x < N, "mapToPointFT: invalid field element");
+        uint256 x = _x;
+        bool decision = isNonResidueFP(x);
```
ChihChengLiang, line 141: It looks like we can do isNonResidueFP in modexp optimization too

contracts/libs/BLS.sol
```diff
@@ -410,24 +324,28 @@ library BLS {
         return !isNonResidue;
     }
 
-    function isNonResidueFP2(uint256[2] memory e)
+    function sqrtFaster(uint256 xx)
         internal
         view
```
ChihChengLiang, line 329: This function is pure
```suggestion
        pure
```

contracts/libs/BLS.sol
```diff
@@ -442,25 +360,29 @@ library BLS {
...
+            x := mload(freemem)
+            hasRoot := eq(xx, mulmod(x, x, N))
         }
-        require(callSuccess, "BLS: isNonResidueFP2 modexp call failed");
-        return !isNonResidue;
+        require(callSuccess, "BLS: sqrt modexp call failed");
     }
 
-    function sqrt(uint256 xx) internal view returns (uint256 x, bool hasRoot) {
+    function inverseFaster(uint256 a) internal view returns (uint256) {
```
ChihChengLiang, line 369: This function is pure too
```suggestion
    function inverseFaster(uint256 a) internal pure returns (uint256) {
```

ts/hash_to_field.ts
```diff
@@ -0,0 +1,209 @@
...
+    const msg = "0x616263";
+    mcl.setDomain(dst);
+    const p = mcl.hashToPoint(msg);
+    const result = mcl.g1ToHex(p);
+    assert.equal(result[0], expected0);
+    assert.equal(result[1], expected1);
+}
+
+import * as mcl from "./mcl";
+async function test() {
```
ChihChengLiang, line 201: Can we organize these test related functions to the `test/`?

ChihChengLiang (commented):
My first pass of review

ts/hash_to_field.ts
kilic: those tests test tests :). 
I've left them for reviewers or auditors to prove that we are in line with the standart. These test don't need be run repeatedly.

ChihChengLiang: I see. I'll later organize them to #202.

ChihChengLiang (approved):
Looks good to me

PR #203, opened 2020-09-07
Remove expectedTxHashCommitment check

ChihChengLiang:
We already check the hash of the whole commitment, no need to check the hash of the committed txs again.

### DIFF
--- a/contracts/libs/Types.sol
+++ b/contracts/libs/Types.sol
@@ -36,8 +36,8 @@ library Types {
     struct Commitment {
         bytes32 stateRoot;
         bytes32 accountRoot;
-        bytes txs;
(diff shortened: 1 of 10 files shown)


vaibhavchellani (approved):
LGTM

contracts/libs/Types.sol
```diff
@@ -36,7 +36,7 @@ library Types {
     struct Commitment {
         bytes32 stateRoot;
         bytes32 accountRoot;
-        bytes32 txHashCommitment;
+        bytes txs;
```
kilic, line 39: moving bytes txs below to signature could be better practice in order not to disturb properties with full evm sizes

Issue #208, opened 2020-09-12
Support Floating Points for compressed transfer amount and fee


ChihChengLiang:
We want to support sending and storing fractions of tokens.

We represent amount and fee fields in floating-point fields

## Stage 1: Hardcoded exponent bits

A field is a 2 bytes data where we use the first 4 bits to represent the exponent of 10, and the rest 12 bits for the mantissa.

```
[exponent_bits<4>|mantisa_bits<12>]
```

This decompresses to number `mantissa * 10^exponent`

### How do I represent number 5?

It's 5 * 10^0
```
[exponent_bits<4>|mantisa_bits<12>]
        0000      0000 0000 0101
```

### How do I represent 5 USDT?

The USDT is an ERC20 token with decimal 6.

That means 5 USDT is 5,000,000 in uint256, on the mainnet contract.

### What's the largest possible amount of USDT I can send in one transaction?

4095 * 10^9 USDT

```
[exponent_bits<4>|mantisa_bits<12>]
        1111      1111 1111 1111
```

### What's the tiniest possible amount of USDT I can send in one transaction?

1 * 10^(-6) USDT

```
[exponent_bits<4>|mantisa_bits<12>]
        0001      0000 0000 0001
```

### If Alice has 5 USDT in her state leaf, what's her uint256 balance field looks like?

5,000,000

### How do I represent 5 Ether?

We represent Ether in Gwei, so 5 Ether would be

5 * 10^9

```
[exponent_bits<4>|mantisa_bits<12>]
        1001      0000 0000 0101
```

### What's the tiniest possible amount of Ether I can send in one transaction?

1 Gwei

### What's the largest possible amount of Ether I can send in one transaction?

4095 * 10^6 Ether

### How do we compress a uint256 value

```python
def compress(input: int) -> int:
    exponent = 0
    mantissa = input
    for _ in range(15):
        if mantissa % 10 == 0:
            mantissa = mantissa /10
            exponent += 1
        else:
            break
    if mantissa > 0x0fff:
        raise ValueError("Invalid input")
    return exponent << 12 + mantissa
```

### How do we decompress a floating-point field

```python
def decompress(input):
    exponent = 0xf000 & input
    mantissa = 0x0fff & input
    return mantissa * 10 ** (exponent)
```

## Stage 2: Allow coordinator to commit the exponent bit.

For example, if the coordinator commits 6, it means we have 6 bits for the exponent of 10 and 10 bits for the mantissa.

```
[...]

PR #211, opened 2020-09-14
Implement floating point

ChihChengLiang:
Part of #208


ChihChengLiang:
Submit batch cost is now brought down from 519144 to 452234 because the amount and fees are reduced from 4 bytes to 2 bytes.

vaibhavchellani (approved):
LGTM

ChihChengLiang:
Thanks for the review!

PR #222, opened 2020-09-16
Optimize submit batch

ChihChengLiang:
```
$npx buidler run scripts/benchmark.ts
submitTransferBatch: Gas cost execution cost for 32 x 32 txs 519084
```

### DIFF
--- a/ts/commitments.ts
+++ b/ts/commitments.ts
@@ -1,4 +1,5 @@
 import { BigNumberish, BytesLike, ethers } from "ethers";
+import { Rollup } from "../types/ethers-contracts/Rollup";
 import { ZERO_BYTES32 } from "./constants";
 import { Hasher, Tree } from "./tree";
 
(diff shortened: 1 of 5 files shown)


ts/commitments.ts
```diff
@@ -202,3 +210,36 @@ export class CommitmentTree {
...
+    async submit(rollup: Rollup, stakingAmount: string) {
+        return await rollup.submitTransferBatch(
+            this.commitments.map(c => c.stateRoot),
+            this.commitments.map(c => c.signature),
+            this.commitments.map(c => c.tokenType),
+            this.commitments.map(c => c.feeReceiver),
+            this.commitments.map(c => c.txs),
+            { value: ethers.utils.parseEther(stakingAmount) }
+        );
+    }
```
ChihChengLiang, line 228: The deconstruct is so ugly so I move the contract method here.

PR #224, opened 2020-09-17
Add simple continuous benchmark steps in CI

ChihChengLiang:
With this PR we can compare the PR performance to the master branch performance so that we can be more aware of the degrade of gas cost.

Screenshot from the [previous run](
https://github.com/ChihChengLiang/RedditHubble/pull/3/checks?check_run_id=1130020872) on my repo 

Screen Shot 2020-09-18 at 2 00 31 AM


vaibhavchellani (approved):
LGTM

PR #225, opened 2020-09-18
remove Tx.sol functions we no longer use

ChihChengLiang:


vaibhavchellani (approved):
Much cleanup, such wow

PR #226, opened 2020-09-18
refactored utils

vaibhavchellani:
Typechain didnt support `structs` as args before so we had to write extra utils for off-chain testing. This PR removes those and creates `ClientUtils.sol` that acts as a frontend for utils. Also converts all functions in `RollupUtils.sol` to internal, which means lower gas costs.


vaibhavchellani:
Closing in favour of #227

PR #227, opened 2020-09-18
abstract out RollupUtilsLib

ChihChengLiang:
This is an alternative to #226

In this PR
- We still keep RollupUtils as the client proxy contract. 
- RollupUtils is now a contract, and 
- We have RollupUtilsLib that defines internal functions for other contracts to consume. This has the benefit that we have lower gas cost and we don't have to link the library for rollupUtils.

### DIFF
--- a/contracts/MassMigrations.sol
+++ b/contracts/MassMigrations.sol
@@ -3,7 +3,7 @@ pragma experimental ABIEncoderV2;
 
 import { FraudProofHelpers } from "./FraudProof.sol";
 import { Types } from "./libs/Types.sol";
-import { RollupUtils } from "./libs/RollupUtils.sol";
+import { RollupUtilsLib } from "./libs/RollupUtils.sol";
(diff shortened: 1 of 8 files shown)


contracts/MassMigrations.sol
```diff
@@ -101,7 +101,7 @@ contract MassMigration is FraudProofHelpers {
         require(
             MerkleTreeUtilsLib.verifyLeaf(
                 stateRoot,
-                RollupUtils.HashFromAccount(fromAccountProof.account),
+                RollupUtilsLib.HashFromAccount(fromAccountProof.account),
```
ChihChengLiang, line 104: Actually `keccak256(RollupUtilsLib.BytesFromAccount(new_account))` might be slightly preferred here. 
- Pros: HashFromAccount is more readable
- Cons: HashFromAccount has two function jumps. 

So I'm delaying the optimization later.

PR #232, opened 2020-09-22
Rename Accounts to States

ChihChengLiang:
After this PR, the term "account" means either a public tree leaf or an Ethereum account in our codebase. No more confusion with the term "state"

### DIFF
--- a/contracts/FraudProof.sol
+++ b/contracts/FraudProof.sol
@@ -24,7 +24,7 @@ contract FraudProofHelpers is FraudProofSetup {
     function validateTxBasic(
         uint256 amount,
         uint256 fee,
-        Types.UserState memory _from_account
+        Types.UserState memory fromState
     ) public pure returns (Types.ErrorCode) {
         if (amount == 0) {
             // invalid state transition
@@ -34,7 +34,7 @@ contract FraudProofHelpers is FraudProofSetup {
         }
 
         // check from leaf has enough balance
-        if (_from_account.balance < amount.add(fee)) {
+        if (fromState.balance < amount.add(fee)) {
             // invalid state transition
             // needs to be slashed because the account doesnt have enough balance
             // for the transfer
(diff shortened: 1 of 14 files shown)


contracts/FraudProof.sol
```diff
@@ -35,69 +34,13 @@ contract FraudProofHelpers is FraudProofSetup {
...
-            // needs to be slashed because the account doesnt have enough balance
+            // needs to be slashed because the state doesnt have enough balance
             // for the transfer
             return Types.ErrorCode.NotEnoughTokenBalance;
         }
 
         return Types.ErrorCode.NoError;
     }
-
-    function RemoveTokensFromAccount(
```
ChihChengLiang, line 48: We removed these functions: RemoveTokensFromAccount, UpdateBalanceInAccount, AddTokensToAccount, and BalanceFromAccount.

contracts/FraudProof.sol
```diff
@@ -35,69 +34,13 @@ contract FraudProofHelpers is FraudProofSetup {
...
-        pure
-        returns (uint256)
-    {
-        return account.balance;
-    }
-
-    /**
-     * @notice Returns the updated root and balance
-     */
-    function UpdateAccountWithSiblings(
```
ChihChengLiang, line 92: We no longer use this either, since this function depends on `_merkle_proof.pathToAccount`.

contracts/libs/Types.sol
```diff
@@ -157,19 +157,39 @@ library Types {
...
 
-    // UserAccount contains the actual data stored in the leaf of balance tree
-    struct UserAccount {
-        // ID is the path to the pubkey in the PDA tree
-        uint256 ID;
+    /**
+    @param pubkeyIndex path to the pubkey in the PDA tree
+     */
+    struct UserState {
+        uint256 pubkeyIndex;
```
ChihChengLiang, line 164: UserAccount is renamed to UserState, and ID is renamed to pubkeyIndex

contracts/libs/Types.sol
```diff
@@ -157,19 +157,39 @@ library Types {
...
         uint256 tokenType;
         uint256 balance;
         uint256 nonce;
     }
 
-    struct AccountMerkleProof {
-        UserAccount account;
-        uint256 pathToAccount; // This field is kept for backward competibility, don't use it.
-        bytes32[] siblings;
+    function encode(UserState memory state)
```
ChihChengLiang, line 170: We just use `state.encode()` everywhere. We use `keccak256(state.encode())` but not `state.hash()` so we don't either use 2 function jumps or rewrite abi.encode twice.

contracts/libs/RollupUtils.sol
```diff
@@ -4,33 +4,9 @@ pragma experimental ABIEncoderV2;
 import { Tx } from "./Tx.sol";
 import { Types } from "./Types.sol";
 
-library RollupUtilsLib {
```
ChihChengLiang, line 7: RollupUtilsLib is removed

ChihChengLiang (commented):
Some quick highlights of the changes

vaibhavchellani (approved):
LGTM

Issue #230, opened 2020-09-21
Dispute Signature testcase


vaibhavchellani:
We are probably signing on the wrong message in the test suite while submitting a batch, we need a testcase to call `disputeSignature` before sending a batch.

vaibhavchellani:
It seems like we are testing the signatures in the testcase by calling `check_signature` [here](https://github.com/thehubbleproject/RedditHubble/blob/master/test/rollupTransfer.test.ts#L118).

However, the [message signed in the test case](https://github.com/thehubbleproject/RedditHubble/blob/master/ts/tx.ts#L58) seems like is not the same as what the [util function will produce](https://github.com/thehubbleproject/RedditHubble/blob/master/contracts/libs/RollupUtils.sol#L152)

PR #233, opened 2020-09-22
Simplify check sig API

ChihChengLiang:
Also, attempt to address #230


ChihChengLiang:
Gas cost of the dispute signature
### On master
- operation gas cost: 4525438
- transaction gas cost: 6256101

### On this PR

- operation gas cost: 4659600
- transaction gas cost: 6389871

vaibhavchellani:
I think the gas costs are okay and we should go ahead with this change for sure!

PR #237, opened 2020-09-24
Build test factories

ChihChengLiang:
So we can build desired states, txs, or commitments without code duplication and duplicated works.
This helps us writing more tests.

### DIFF
--- a/ts/stateTree.ts
+++ b/ts/stateTree.ts
@@ -60,9 +60,17 @@ export class StateTree {
         this.stateTree.updateSingle(stateID, leaf);
         this.states[stateID] = state;
     }
+    public createStateBulk(states: State[]) {
+        for (const state of states) {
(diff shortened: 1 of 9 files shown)


ts/stateTree.ts
```diff
@@ -58,12 +58,21 @@ export class StateTree {
         }
         const leaf = state.toStateLeaf();
         this.stateTree.updateSingle(stateID, leaf);
-        this.states[stateID] = state;
+        // Need to clone the object so that whatever we do on this.states later won't affect the input state.
+        this.states[stateID] = state.clone();
```
ChihChengLiang, line 62: Fixed a gotcha here. Before this change, the input state would change when we run state transition on states

vaibhavchellani (approved):
Such cleanup, much wow.

PR #239, opened 2020-09-25
remove production contracts

ChihChengLiang:
We created XProduction contracts because the fraud proof contracts depend on too many external contracts and their constructors were complicated. It was difficult to do unit tests for the fraud proof contracts. So we moved the constructor to an inherited contract named Xproduction.

But now our fraudproof contracts no longer depend on many other contracts. No complexity in unit tests. The Xproduction contracts can go.

### DIFF
--- a/contracts/prod/MassMigration.sol
+++ b/contracts/prod/MassMigration.sol
@@ -1,6 +1,0 @@
-pragma solidity ^0.5.15;
-pragma experimental ABIEncoderV2;
-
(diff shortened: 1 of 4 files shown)


vaibhavchellani (approved):
Can you point out the reason for change, otherwise looks good to me!

ChihChengLiang:
Sure thing, I've updated on the pull request description.

PR #240, opened 2020-09-25
Slither clean

ChihChengLiang:
Added some scripts and configs to run slither

Install slither with 

```
pip3 install slither-analyzer
```

Run slither with 
```
npm run slither
```


contracts/interfaces/IERC20.sol
```diff
@@ -1,14 +0,0 @@
-pragma solidity ^0.5.15;
-
-// ERC20 token interface
-contract IERC20 {
```
ChihChengLiang, line 4: @vaibhavchellani Why we use the customized IERC20 interface but not the openzepplin's?

vaibhavchellani: No reason should switch to openzepplin.

PR #242, opened 2020-09-26
Rework BLS ts API

ChihChengLiang:
Cleaned up the ts API, so that in most of the case we just handle with the solidity structure without worrying about the mcl conversion.

### DIFF
--- a/test/rollupTransfer.test.ts
+++ b/test/rollupTransfer.test.ts
@@ -39,7 +39,7 @@ describe("Rollup Transfer Commitment", () => {
         registry = await AccountRegistry.new(registryContract);
         states = UserStateFactory.buildList(STATE_SIZE);
         for (const state of states) {
-            await registry.register(state.encodePubkey());
+            await registry.register(state.getPubkey());
         }
     });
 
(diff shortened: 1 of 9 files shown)


ts/mcl.ts
```diff
@@ -28,132 +39,120 @@ export function setDomainHex(domain: string) {
...
-    _acc.normalize();
-    return _acc;
+export function aggreagate(signatures: Signature[]): solG1 {
+    let aggregated = new mcl.G1();
+    for (const sig of signatures) {
+        aggregated = mcl.add(aggregated, sig);
+    }
+    aggregated.normalize();
+    return g1ToHex(aggregated);
 }
```
ChihChengLiang, line 129: Change the aggregate to aggregate only the signatures. We are not aggregating anything else.

test/rollupTransfer.test.ts
```diff
@@ -52,18 +52,17 @@ describe("Rollup Transfer Commitment", () => {
...
             const sender = states[tx.fromIndex];
-            pubkeys.push(sender.encodePubkey());
+            pubkeys.push(sender.getPubkey());
             pubkeyWitnesses.push(registry.witness(sender.pubkeyIndex));
-            const signature = sender.sign(tx);
-            aggSignature = mcl.aggreagate(aggSignature, signature);
+            signatures.push(sender.sign(tx));
         }
-        const signature = mcl.g1ToHex(aggSignature);
+        const signature = mcl.aggreagate(signatures)
```
ChihChengLiang, line 65: instead of initiating a G1, we just aggregate an array of signatures.

vaibhavchellani (approved):
LGTM

PR #243, opened 2020-09-26
Minor fix in Account trees

ChihChengLiang:
Mostly cosmetics

- use callStatic for the result and the gas operation cost
- use randomLeaves for leave generaton


vaibhavchellani (approved):
LGTM

PR #252, opened 2020-09-29
add validations in ClaimTokens

ChihChengLiang:
TODO: add some tests

### DIFF
--- a/contracts/WithdrawManager.sol
+++ b/contracts/WithdrawManager.sol
@@ -7,6 +7,9 @@ import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { Tx } from "./libs/Tx.sol";
 import { NameRegistry as Registry } from "./NameRegistry.sol";
 import { Vault } from "./Vault.sol";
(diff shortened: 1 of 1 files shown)


contracts/WithdrawManager.sol
```diff
@@ -59,12 +74,27 @@ contract WithdrawManager {
...
+            accountRegistry.root(),
+            keccak256(abi.encodePacked(pubkey)),
+            accountID,
+            witness
+        );
+        BLS.verifySingle(
+            signature,
+            pubkey,
+            BLS.hashToPoint(APP_ID, abi.encodePacked(msg.sender))
+        );
```
ChihChengLiang, line 94: The validation has these properties:
- The pubkey must be the one associated with the accountID
- Must be signed against the same pubkey.
- The message contains `msg.sender` so that frontrunner can't steal the claim.

vaibhavchellani (approved):
LGTM

ChihChengLiang:
Thanks for the review, I think I'll add tests in the next pull request. Created an issue here #253

PR #231, opened 2020-09-22
added create2transfer

vaibhavchellani:

PR #257, opened 2020-09-30
Fixing mass migration

ChihChengLiang:
Some summary of the discussion today:

- no point checking spokeID in state transition, we should check it in dispute signature
- no point checking BURN_STATE_INDEX in the state transition, since we don't modify the receiver state
- no point checking tx.tokenType all equal, check them against the committed tokenType instead.

### DIFF
--- a/contracts/libs/Tx.sol
+++ b/contracts/libs/Tx.sol
@@ -17,7 +17,6 @@ library Tx {
     uint256 public constant MANTISSA_BITS = 12;
     uint256 public constant MASK_MANTISSA = 0x0fff;
     uint256 public constant MASK_NONCE = 0xffffffff;
-    uint256 public constant MASK_SPOKE = 0xffffffff;
     uint256 public constant MASK_TOKEN_ID = 0xffff;
     uint256 public constant MASK_BYTE = 0xff;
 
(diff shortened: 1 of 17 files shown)


contracts/libs/Tx.sol
```diff
@@ -278,4 +267,20 @@ library Tx {
...
         return txs.length / TX_LEN_5;
     }
+
+    function massMigration_messageOf(
+        MassMigration memory _tx,
+        uint256 nonce,
+        uint256 spokeID
+    ) internal pure returns (bytes memory) {
+        return
+            abi.encodePacked(
```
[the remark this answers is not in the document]
ChihChengLiang: I am now more convinced that we should use `encodePack` for the messages. The reason is when we do hash to field in BLS, there's a loop that iterates every 32 bytes we added to the raw message
https://github.com/thehubbleproject/hubble-contracts/blob/master/contracts/libs/BLS.sol#L338-L344

ts/tx.ts
```diff
@@ -127,15 +127,17 @@ export class TxMassMigration implements SignableTx {
...
-            this.TX_TYPE,
-            hexZeroPad(hexlify(this.fromIndex), stateIDLen),
-            hexZeroPad(this.amount.toHexString(), 32),
-            hexZeroPad(this.fee.toHexString(), 32),
-            hexZeroPad(hexlify(this.nonce), nonceLen),
-            hexZeroPad(hexlify(this.spokeID), spokeLen)
-        ]);
-        return hexlify(concated);
+        return solidityPack(
+            ["uint8", "uint32", "uint256", "uint256", "uint32", "uint32"],
```
vaibhavchellani, line 131: from index is `uint32` and to index is `uint256` not sure if it is a typo, just pointing out.

ChihChengLiang: Oh, that's not a toIndex, we don't have the toIndex now.

vaibhavchellani: Ah! Right, gotcha.

vaibhavchellani (approved):
looks good to me, might have some conflicts to resolve post merging #231

PR #269, opened 2020-10-07
Is last leaf

### LINKED ISSUE #265: Attackers can slash the first commitment in the batch with previous commit
### What's wrong

To dispute the first commitment in the batch, we require the challenger to provide the last commit in the previous batch. However, we currently only require that commit to be in the previous batch but have no way to make sure that's the last commit.

So if the previous batch has 3 commits, an attacker can provide the second commit with proofs to trick the dispute function that the target commit is fraudulent.  

https://github.com/thehubbleproject/hubble-contracts/blob/c0228fbda6951257cb27b84dbd9ca81708491703/contracts/Rollup.sol#L105-L110

### How could we fix it
We need to commit the number of commits in a batch somewhere so that we can check if a previous commit is a last commit in the batch.

ChihChengLiang:
Fix #265

### DIFF
--- a/contracts/MerkleTreeUtils.sol
+++ b/contracts/MerkleTreeUtils.sol
@@ -173,4 +173,28 @@ contract MerkleTreeUtils {
     ) public pure returns (bytes32) {
         return MerkleTreeUtilsLib.rootFromWitnesses(_leaf, _path, _siblings);
     }
+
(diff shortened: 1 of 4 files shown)


ChihChengLiang:
close in favor of  #277 and #278

PR #271, opened 2020-10-10
Fix submit deposit batch

### LINKED ISSUE #181: Slashing process removes ambiguous deposit root
In rollback process we iterate back to the challenge point. In this iteration deposit manager is called to enqueue a deposit subtree even if this batch is not a deposit batch. 

https://github.com/thehubbleproject/RedditHubble/blob/59affd3e26ac8c2f9c9ff945ad3b376e9e3d1c68/contracts/rollup.sol#L146

This can be handled with a workaround by disabling processing deposit roots with zero.

However there might better solution:

* Drop using `depositRoot` field instead use `commitmentRoot` for deposit roots.
* During finalising a queue mark finalised queue (if marked it means already submitted)
* Call deposit manager in the rollback iteration
* Check if queue submitted if so remove the mark

ChihChengLiang:
Fix  #181 and some other stuff.

DepositManager now does two things:

- dequeue a subtree and mark it submitted at batchID
- enqueue a subtree if it has been submitted at batchID before

Since DepositManager already trusts Rollup contract, so we move subtree vacancy check to the Rollup contract.

This solves the problem we had:
1. We've been submitting empty deposit roots in every non-deposit Batch. Now we don't need a field in Batch.
2. When rollback the batches, we are enqueuing zero bytes to the queue.

### DIFF
--- a/contracts/DepositManager.sol
+++ b/contracts/DepositManager.sol
@@ -2,7 +2,6 @@ pragma solidity ^0.5.15;
 pragma experimental ABIEncoderV2;
 import { Types } from "./libs/Types.sol";
 import { Logger } from "./Logger.sol";
-import { MerkleTreeUtils as MTUtils } from "./MerkleTreeUtils.sol";
 import { NameRegistry as Registry } from "./NameRegistry.sol";
 import { ITokenRegistry } from "./interfaces/ITokenRegistry.sol";
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
@@ -91,7 +90,6 @@ contract DepositManager is DepositCore {
     Registry public nameRegistry;
     address public vault;
 
-    MTUtils public merkleUtils;
     Governance public governance;
     Logger public logger;
     ITokenRegistry public tokenRegistry;
@@ -120,9 +118,6 @@ contract DepositManager is DepositCore {
         nameRegistry = Registry(_registryAddr);
         governance = Governance(
             nameRegistry.getContractDetails(ParamManager.Governance())
-        );
-        merkleUtils = MTUtils(
-            nameRegistry.getContractDetails(ParamManager.MERKLE_UTILS())
         );
         tokenRegistry = ITokenRegistry(
             nameRegistry.getContractDetails(ParamManager.TOKEN_REGISTRY())
--- a/contracts/DepositManager.sol
+++ b/contracts/DepositManager.sol
@@ -183,6 +183,7 @@ contract DepositManager is DepositCore {
         bytes32 subtreeRoot = submittedSubtree[batchID];
         if (subtreeRoot != bytes32(0)) {
             enqueue(subtreeRoot);
+            delete submittedSubtree[batchID];
         }
     }
 }
--- a/contracts/DepositManager.sol
+++ b/contracts/DepositManager.sol
@@ -171,38 +171,12 @@ contract DepositManager is DepositCore {
         }
     }
 
-    /**
-     * @notice Merges the deposit tree with the balance tree by
-     *        superimposing the deposit subtree on the balance tree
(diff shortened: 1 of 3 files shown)


contracts/DepositManager.sol
```diff
@@ -2,7 +2,6 @@ pragma solidity ^0.5.15;
 pragma experimental ABIEncoderV2;
 import { Types } from "./libs/Types.sol";
 import { Logger } from "./Logger.sol";
-import { MerkleTreeUtils as MTUtils } from "./MerkleTreeUtils.sol";
```
ChihChengLiang, line 5: No merkle utils in DepositManager now 🎉

contracts/DepositManager.sol
```diff
@@ -171,41 +169,21 @@ contract DepositManager is DepositCore {
...
     }
 
-    /**
-     * @notice Merges the deposit tree with the balance tree by
-     *        superimposing the deposit subtree on the balance tree
-     * @param _subTreeDepth Deposit tree depth or depth of subtree that is being deposited
-     * @param zero Merkle proof proving the node at which we are inserting the deposit subtree consists of all empty leaves
-     * @return Updates in-state merkle tree root
-     */
-    function finaliseDeposits(
```
ChihChengLiang, line 181: The verb finalise for deposits should be avoided because it implies nothing will ever change again. But chances are the submitted deposits might be enqueued back.

contracts/Rollup.sol
```diff
@@ -124,13 +124,6 @@ contract RollupSetup {
 }
 
 contract RollupHelpers is RollupSetup {
-    /**
-     * @notice Returns the latest state root
-     */
-    function getLatestBalanceTreeRoot() public view returns (bytes32) {
-        return batches[batches.length - 1].commitmentRoot;
-    }
```
ChihChengLiang, line 132: This API says get BalanceTreeRoot but gives commitmentRoot lol

contracts/Rollup.sol
```diff
@@ -406,19 +397,34 @@ contract Rollup is RollupHelpers {
...
-        uint256 _subTreeDepth,
-        Types.StateMerkleProofWithPath calldata zero
-    ) external payable onlyCoordinator isNotRollingBack {
-        bytes32 depositSubTreeRoot = depositManager.finaliseDeposits(
-            _subTreeDepth,
-            zero,
-            getLatestBalanceTreeRoot()
+        Types.CommitmentInclusionProof memory previous,
+        Types.SubtreeVacancyProof memory vacant
+    ) public payable onlyCoordinator isNotRollingBack {
```
ChihChengLiang, line 402: Solidity complains `Umimplemented Error` when we use calldata structs here, so I change the visibility to `public` here. 
In contrast to the case of submit transfer batch, where we want to send as much data and be gas efficient there, the submit deposit batch doesn't have the same requirement here. wdyt @vaibhavchellani

contracts/libs/Types.sol
```diff
@@ -35,7 +35,6 @@ library Types {
         bytes32 commitmentRoot;
         address committer;
         uint256 finalisesOn;
-        bytes32 depositRoot;
```
ChihChengLiang, line 38: depositRoot is gone now.

contracts/DepositManager.sol
vaibhavchellani: Wohoo!

contracts/DepositManager.sol
vaibhavchellani: Yeah that makes sense to me

contracts/Rollup.sol
vaibhavchellani: Yeah I think that makes sense.

vaibhavchellani (approved):
Lets merge! Amazing job @ChihChengLiang

PR #273, opened 2020-10-10
Rollup nitpick round

ChihChengLiang:


vaibhavchellani (approved):
LGTM

test/rollup.test.ts
```diff
@@ -118,7 +118,7 @@ describe("Rollup", async function() {
         const previousMP = initialBatch.proofCompressed(0);
         const commitmentMP = targetBatch.proof(0);
 
-        const _tx = await rollup.disputeBatch(
+        const _tx = await rollup.disputeTransition_Transfer(
```
vaibhavchellani, line 121: Do we want to stick with camel case?

ChihChengLiang: Hmmm, the original idea was to get the clarity by splitting the feature "Transfer" from the nature of the dispute, "a dispute of state transition".
But maybe following the mixedCase for function as style guide suggested for simplicity is better. Will fix this

ChihChengLiang: Fixed

PR #276, opened 2020-10-11
Batch meta

ChihChengLiang:


contracts/Rollup.sol
```diff
@@ -247,34 +248,44 @@ contract Rollup is RollupHelpers {
...
-                genesisCommitments
-            ),
-            committer: msg.sender,
-            finalisesOn: block.number + governance.TIME_TO_FINALISE()
+            commitmentRoot: commitmentRoot,
+            meta: Types.encodeMeta(
+                uint256(Types.Usage.Genesis),
+                1,
+                msg.sender,
+                block.number // genesis finalise instantly
```
ChihChengLiang, line 264: Headsup here. I let the genesis finalise instantly here because we are unlikely to dispute it.

vaibhavchellani: Yeah that sounds reasonable

vaibhavchellani (approved):
LGTM

PR #277, opened 2020-10-11
Revamp merkle utils

ChihChengLiang:
- Average cost of verifying a leaf 14892.4
- Merklizing 32 leaves on chain 18180


vaibhavchellani (approved):
Love it!

PR #281, opened 2020-10-12
Solhint

ChihChengLiang:


.solhint.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "solhint:recommended",
+  "rules": {
+    "reason-string": ["warn",{"maxLength":80}],
+    "var-name-mixedcase": ["error"],
+    "func-name-mixedcase": ["error"],
```
ChihChengLiang, line 6: Try to enforce these two extra style rules

contracts/Rollup.sol
```diff
@@ -216,36 +216,36 @@ contract Rollup is RollupHelpers {
     bytes32
         public constant ZERO_BYTES32 = 0x290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563;
 
-    bytes32 public APP_ID;
+    bytes32 public appID;
```
ChihChengLiang, line 219: If we upgrade to 0.6 we can define it as immutable

contracts/libs/ModExp.sol
```diff
@@ -1,6 +1,11 @@
 pragma solidity ^0.5.15;
 
-library modexp_3064_fd54 {
+/**
+    @notice Compute $input^(N - 2) mod N$ using Addition Chain method.
+    Where     N = 0x30644e72e131a029b85045b68181585d97816a916871ca8d3c208c16d87cfd47
+    and   N - 2 = 0x30644e72e131a029b85045b68181585d97816a916871ca8d3c208c16d87cfd45
```
ChihChengLiang, line 6: Almost forgot what this number is. It's hard to deduct from the library name, so adding doc strings here.

contracts/libs/ParamManager.sol
```diff
@@ -1,69 +1,69 @@
...
         return keccak256("param_manager");
     }
 
-    function TOKEN_REGISTRY() public pure returns (bytes32) {
+    function tokenRegistry() public pure returns (bytes32) {
         return keccak256("token_registry");
     }
 
-    function TRANSFER() public pure returns (bytes32) {
+    function transferSimple() public pure returns (bytes32) {
```
ChihChengLiang, line 48: We don't use the name `transfer` because otherwise solhint would think we are doing a token transfer and warns reentry risk.

contracts/DepositManager.sol
```diff
@@ -36,7 +36,7 @@ contract DepositCore is SubtreeQueue {
     bytes32[] public babyTrees;
     uint256 public depositCount = 0;
 
-    uint256 public MAX_SUBTREE_SIZE;
+    uint256 public govMaxSubtreeSize;
```
ChihChengLiang, line 39: Before this rename, I often forgot maxSubtreeSize is not a constant and assume it is.

vaibhavchellani: Same

vaibhavchellani (approved):
Will have to deal with more warnings now :))

PR #282, opened 2020-10-12
Client frontend

ChihChengLiang:


contracts/client/ClientFrontend.sol
```diff
@@ -0,0 +1,113 @@
...
+            newRoot,
+            _tx.toIndex,
+            _tx.amount,
+            tokenType,
+            to
+        );
+        return (newRoot, senderState, receiverState, result);
+    }
+
+    function applyTransferTx(
```
ChihChengLiang, line 78: @vaibhavchellani  Does the go client require state root to be exposed in this API? [This line](https://github.com/thehubbleproject/reddit-commander/blob/ad827761d50f9b7989dc22286f18163ec858e204/core/tx.go#L92) suggests it doesn't, but not sure if I'm checking the right line

vaibhavchellani: Ideally I would want it to return the newState after apply, I am not sure about the purpose right now but I can see a usecase where a user would want `fast-sync`, and during that we can just apply transactions without processTx and just validate state updates via stateRoot check. 

But I think we can skip for now and add that later.

contracts/client/Offchain.sol
```diff
@@ -0,0 +1,89 @@
...
+    struct Transfer {
+        uint256 txType;
+        uint256 fromIndex;
+        uint256 toIndex;
+        uint256 amount;
+        uint256 fee;
+        uint256 nonce;
+    }
+
+    function decodeTransfer(bytes memory txBytes)
```
ChihChengLiang, line 16: Should check if the amount and fee passed here are compressible. Will think about it more in the next pr

contracts/client/ClientFrontend.sol
```diff
@@ -0,0 +1,97 @@
...
+import { Types } from "../libs/Types.sol";
+import { Transition } from "../libs/Transition.sol";
+import { Offchain } from "./Offchain.sol";
+
+contract ClientFrontend {
+    using SafeMath for uint256;
+    using Tx for bytes;
+    using Types for Types.UserState;
+
+    function decodeTransfer(bytes calldata txBytes)
```
vaibhavchellani, line 15: all the variables named txBytes are encodedTxBytes correct? A small nitpick would be to change the variable name to reflect that, but you can ignore that if you don't feel a need for it

vaibhavchellani (approved):
LGTM!

contracts/client/ClientFrontend.sol
ChihChengLiang: Yes, they are encoded txBytes. I can rename those

PR #283, opened 2020-10-13
Transition: cleaner state transition logic

ChihChengLiang:
We now have every state transition fits the following style

```
 processSender
  validateSender
  applySender
processReceiver
  validateReceiver
  applyReceiver
```
And we can remove all processFees, because they are just another processReceiver.

### DIFF
--- a/contracts/Transfer.sol
+++ b/contracts/Transfer.sol
@@ -88,19 +88,17 @@ contract Transfer {
                 proofs[i * 2],
                 proofs[i * 2 + 1]
             );
-            if (result != Types.Result.Ok) break;
+            if (result != Types.Result.Ok) return (stateRoot, result);
             // Only trust fees when the result is good
             fees = fees.add(_tx.fee);
         }
-        if (result == Types.Result.Ok) {
-            (stateRoot, , result) = Transition.processReceiver(
-                stateRoot,
-                feeReceiver,
-                fees,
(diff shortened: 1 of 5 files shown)


contracts/Transfer.sol
```diff
@@ -128,111 +125,29 @@ contract Transfer {
...
+            from
         );
         if (result != Types.Result.Ok) return (bytes32(0), "", "", result);
-
-        if (from.state.tokenType != tokenType) {
-            return (bytes32(0), "", "", Types.Result.BadFromTokenType);
-        }
-
-        if (to.state.tokenType != tokenType)
-            return (bytes32(0), "", "", Types.Result.BadToTokenType);
```
ChihChengLiang, line 156: We used to have this dangerous line that we judge the stuff provided by the disputer without validating it.
The Merkle verify is at the 10 lines below, but here we are already trying to slash the coordinator by whatever states provided by the disputer.

ChihChengLiang:
Gas cost for state transition. I think the reason that costs slightly increased is we have more internal function calls than before.

### Transfer

- before: 2612004
- after: 2612452

### mass migration

- before: 1257996
- after:    1264140

### create2Transfer
no data for state transfer yet.

vaibhavchellani (approved):
Really love the new API's. Excellent work!

PR #288, opened 2020-10-15
Revamp merkle tree again

ChihChengLiang:
- Renames
  - verifyLeaf -> verify
  - getMerkleRootFromLeaves -> merklise
  - MerkleTreeUtils -> MerkleTree
- Move MerkleTree to libs/
- Change behavior
  - We can now only merklise at max 32 leaves to root on chain. These are used in merklising commitments and mass migration withdraw root check.
- Exploit memory
  - Merklising 32 nodes is slightly more expensive 18100 -> 18422
  - Cost of getting default Root for a certain depth is reduced from 1744 to 880


contracts/libs/MerkleTree.sol
```diff
@@ -0,0 +1,101 @@
...
+        hashes[21] = 0xedf260291f734ddac396a956127dde4c34c0cfb8d8052f88ac139658ccf2d507;
+        hashes[22] = 0x6075c657a105351e7f0fce53bc320113324a522e8fd52dc878c762551e01a46e;
+        hashes[23] = 0x6ca6a3f763a9395f7da16014725ca7ee17e4815c0ff8119bf33f273dee11833b;
+        hashes[24] = 0x1c25ef10ffeb3c7d08aa707d17286e0b0d3cbcb50f1bd3b6523b63ba3b52dd0f;
+        hashes[25] = 0xfffc43bd08273ccf135fd3cacbeef055418e09eb728d727c4d5d5c556cdea7e3;
+        hashes[26] = 0xc5ab8111456b1f28f3c7a0a604b4553ce905cb019c463ee159137af83c350b22;
+        hashes[27] = 0x0ff273fcbf4ae0f2bd88d6cf319ff4004f8d7dca70d4ced4e74d2c74139739e6;
+        hashes[28] = 0x7fa06ba11241ddd5efdc65d4e39c9f6991b74fd4b81b62230808216c876f827c;
+        hashes[29] = 0x7e275adf313a996c7e2950cac67caba02a5ff925ebf9906b58949f3e77aec5b9;
+        hashes[30] = 0x8f6162fa308d2b3a15dc33cffac85f13ab349173121645aedf00f471663108be;
```
vaibhavchellani, line 64: Cost of making it a lib I guess? Is there any other way?

vaibhavchellani (approved):
LGTM

contracts/libs/MerkleTree.sol
ChihChengLiang: Yeah, but the cost comes with other benefits like no linking, no external call, and no view storage cost. It's a pill to swallow 💊.

vaibhavchellani: And a tasty one

PR #294, opened 2020-10-19
alternative transition design

ChihChengLiang:
- Highlight the Merkle checks and compute roots in `processX` API.
- Merge `validateX` and `applyX` APIs, because those are what the coordinator and the client care.
- The client consumes `validateAndApply` API.
  - No Merkle computation in those functions.
  - states input/output are in encoded form. The client can be state agnostic.


vaibhavchellani (approved):
Love it!

PR #295, opened 2020-10-20
Refactor packaging

fredlacs:
Refactored the packaging so it can be more easily used as a library.
Also fixed a small typing error for it to compile.

Currently adding this dependency straight from git, but it can be added to the npm registry.


ChihChengLiang:
closing for the same reason here https://github.com/thehubbleproject/hubble-contracts/pull/296#issuecomment-716308917

PR #296, opened 2020-10-20
Browser support

fredlacs:
This PR replaces the Buffer dependency with `vanillajs` when used in the `mcl` library.
This change allows the library to maintain browser compatibility.


ChihChengLiang:
Thank you for the PR, per our offline discussion we now  maintain a general purpose package to support browsers and wallet usage. Please see: https://www.npmjs.com/package/@thehubbleproject/bls

PR #302, opened 2020-10-26
Expose processX api

ChihChengLiang:
1. Moving processX internal function to Transition.sol
2. expose them in frontend contract


vaibhavchellani (approved):
LGTM

PR #308, opened 2020-10-29
Submit commitment check

ChihChengLiang:
1. check if compressed txs are decompressible
2. check if they exceed govMaxTxsPerCommit

Note: Doing these checks in submitBatch might hurt tps. Also, we use 2 function calls, could combine them into 1. I'm thinking of doing this for the time being and optimize later

### DIFF
--- a/contracts/Rollup.sol
+++ b/contracts/Rollup.sol
@@ -287,6 +287,10 @@ contract Rollup is RollupHelpers {
         bytes32 accountRoot = accountRegistry.root();
         bytes32 bodyRoot;
         for (uint256 i = 0; i < stateRoots.length; i++) {
+            require(
+                !txss[i].transferHasExcessData(),
+                "Rollup: transfer has excess data"
+            );
             // This is TransferBody toHash() but we don't want the overhead of struct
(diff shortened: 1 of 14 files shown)


ChihChengLiang:
Master
```
submitTransferBatch: Gas cost execution cost for 32 x 32 txs 404026
Transaction per second 2437.01
```
This PR
```
submitTransferBatch: Gas cost execution cost for 32 x 32 txs 465559
Transaction per second 2114.91
```
Kinda hurt hummmm...

contracts/Rollup.sol
```diff
@@ -299,6 +300,14 @@ contract Rollup is RollupHelpers {
         bytes32 accountRoot = accountRegistry.root();
         bytes32 bodyRoot;
         for (uint256 i = 0; i < stateRoots.length; i++) {
+            require(
```
vaibhavchellani, line 303: Can we instead do this check in fraud-proof and slash appropriately?

ChihChengLiang: Sounds good, will fix.

vaibhavchellani (approved):
LGTM

PR #310, opened 2020-10-29
add missing log new batch

ChihChengLiang:


contracts/Logger.sol
```diff
@@ -6,20 +6,14 @@ contract Logger {
     /*********************
      * Rollup Contract *
      ********************/
-    event NewBatch(
-        address committer,
-        bytes32 updatedRoot,
```
ChihChengLiang, line 11: Removed the updated Root, because we are expected to get
- all submitted state roots from the calldata.
- The genesis root can be accessed through the contructor.

vaibhavchellani: Makes sense to me

contracts/Rollup.sol
```diff
@@ -283,7 +278,9 @@ contract Rollup is RollupHelpers {
             )
         });
         batches.push(newBatch);
-        stake(batches.length - 1);
+        uint256 batchID = batches.length - 1;
+        stake(batchID);
+        logger.logNewBatch(msg.sender, batchID, batchType);
```
vaibhavchellani, line 283: I wonder if not creating a batchID variable is cheaper.

ChihChengLiang: Sounds about right

vaibhavchellani (approved):
LGTM

PR #311, opened 2020-10-29
remove comments

ChihChengLiang:
We can still get good readabilities without those comments


vaibhavchellani (approved):
yeah these were ugly

PR #314, opened 2020-10-30
Rework rollup parameters

ChihChengLiang:
### What's the issue

We have parameters to determine these things: 

1. how many deposits per package we want to insert to the state tree?
2. after how many blocks a batch can be consider finalised?
3. How much gas left is required before you continue rolling back the next batch?
4. How many transactions are allowed to be included in a commit?
5. How much ETH a coordinator has to stake to submit a batch?

But before this PR, we

- make external calls to Governance contract to get the values
- Can't change the parameters

### How is it fixed

- Parameters are natives to the contract. No external call.
- Parameters are controlled by an Ownable contract, we can assign an owner to change it.

Downsides: Not all Parameters are in the same contract. 1 is in deposit manager, 2~5 are in rollup/Parameters.


vaibhavchellani (approved):
LGTM

PR #316, opened 2020-10-31
change asserts to require

ChihChengLiang:
assert takes all gas if failed. In our use cases, it looks unnecessary.


vaibhavchellani (approved):
LGTM

PR #321, opened 2020-11-02
rm gov

ChihChengLiang:
Missed out in #314


ChihChengLiang:
Trivial so merging without a review.

PR #328, opened 2020-11-11
Pubkeys on invalid subgroup checks

kilic:
* Remove `isValidPubKey` check at registry call.
* Add precompile gas cost calculations
    * Send exact amount of gas to precompile calls
* `callSuccess` mostly indicates that one of points (can be G1 or G2) are not on curve or on correct subgroup.
* Batches using Invalid pubkeys can be slashed calling same path with invalid signature aggregation

```
        Types.Result result = transfer.checkSignature(
            target.commitment.body.signature,
            signatureProof,
            target.commitment.stateRoot,
            target.commitment.body.accountRoot,
            appID,
            target.commitment.body.txs
        );

        if (result != Types.Result.Ok) {
            invalidBatchMarker = batchID;
            slashAndRollback();
        }
```

`result` must return __not Ok__ if any of senders pubkey is invalid

## Problem within this PR
* We want to detect that precompile call in another saying the signature verification fails because of a point not being on curve or in correct subgroup. In order to make gas estimation work well for a tx that goes through unsuccesful precompile call, gas amount that is sent along precompile call is hardcoded. Problem is that we before experienced precompile gas cost change with EIP 1108.

### DIFF
--- a/contracts/BLSAccountRegistry.sol
+++ b/contracts/BLSAccountRegistry.sol
@@ -12,10 +12,6 @@ contract BLSAccountRegistry is AccountTree {
     }
 
     function register(uint256[4] calldata pubkey) external returns (uint256) {
-        require(
-            BLS.isValidPublicKey(pubkey),
-            "BLSAccountTree: invalid pub key"
-        );
         bytes32 leaf = keccak256(abi.encodePacked(pubkey));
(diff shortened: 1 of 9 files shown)


ChihChengLiang (approved):
Looks good to me

PR #329, opened 2020-11-11
Upgrade to hardhat

ChihChengLiang:


package.json
```diff
@@ -29,25 +29,22 @@
...
+    "@nomiclabs/hardhat-ethers": "^2.0.0",
     "@openzeppelin/contracts": "^2.4.0",
     "@typechain/ethers-v5": "^1.0.0",
-    "@types/chai": "^4.2.12",
-    "@types/mocha": "^8.0.3",
+    "@types/chai": "^4.2.14",
+    "@types/mocha": "^8.0.4",
     "bn.js": "^5.1.3",
     "chai": "^4.2.0",
-    "ethereum-waffle": "^3.1.0",
```
ChihChengLiang, line 41: Removed Waffle because it depends on ganache, which depends on other horrible things that `npm ci` could not even install. We only use its test helpers, which can be easily replaced with functions from ethers.js.

test/MassMigration.test.ts
```diff
@@ -180,9 +181,10 @@ describe("Mass Migrations", async function() {
 
         const batchId = Number(await rollup.numOfBatchesSubmitted()) - 1;
 
-        await expect(
-            withdrawManager.processWithdrawCommitment(batchId, batch.proof(0))
-        ).revertedWith("Vault: Batch shoould be finalised");
+        await expectRevert(
+            withdrawManager.processWithdrawCommitment(batchId, batch.proof(0)),
+            "Vault: Batch shoould be finalised"
+        );
```
ChihChengLiang, line 187: this is how we expect a transaction to revert

test/deposit.test.ts
```diff
@@ -78,28 +78,37 @@ describe("DepositManager", async function() {
...
-            .and.to.emit(logger, "DepositSubTreeReady")
-            .withArgs(pendingDeposit);
+        assert.equal(event1.args?.pubkeyID.toNumber(), 1);
+        assert.equal(event1.args?.data, deposit1.encode());
+
+        const [eventReady] = await logger.queryFilter(
+            logger.filters.DepositSubTreeReady(null),
+            txDeposit1.blockHash
+        );
+        assert.equal(eventReady.args?.root, pendingDeposit);
```
ChihChengLiang, line 112: This is how we query an event emitted from the logger contract and assert the logs' values.

vaibhavchellani (approved):
Great work! Lets merge!

PR #332, opened 2020-11-13
you can stake more than required

ChihChengLiang:
This main purpose of this change is to make the client dev easier to test a huge amount of staking before getting an exact stake value


vaibhavchellani (approved):
LGTM

PR #337, opened 2020-11-17
Add burn auction

ChihChengLiang:
from https://github.com/iden3/rollup/blob/master/contracts/RollupBurnAuction.sol

### DIFF
--- a/contracts/proposers/BurnAuction.sol
+++ b/contracts/proposers/BurnAuction.sol
@@ -1,8 +1,9 @@
 pragma solidity ^0.5.15;
 import { Chooser } from "./Chooser.sol";
 import { Bitmap } from "../libs/Bitmap.sol";
+import { Ownable } from "@openzeppelin/contracts/ownership/Ownable.sol";
 
-contract BurnAuction {
+contract BurnAuction is Chooser, Ownable {
     uint32 constant BLOCKS_PER_SLOT = 100;
     uint32 constant DELTA_BLOCKS_INITIAL_SLOT = 1000;
 
@@ -24,8 +25,6 @@ contract BurnAuction {
     // slot -> hasProposed
     mapping(uint256 => uint256) public hasProposed;
 
-    address public rollup;
-
     /**
      * @dev Event called when an operator beat the bestBid of the ongoing auction
      */
@@ -35,11 +34,8 @@ contract BurnAuction {
      * @dev RollupBurnAuction constructor
      * Set first block where the slot will begin
      * Initializes auction for first slot
-     * @param _rollup rollup main smart contract address
      */
-    constructor(address _rollup) public {
-        require(_rollup != address(0), "Address 0 inserted");
-        rollup = _rollup;
+    constructor() public {
         genesisBlock = getBlockNumber() + DELTA_BLOCKS_INITIAL_SLOT;
     }
 
@@ -63,8 +59,7 @@ contract BurnAuction {
         emit newBestBid(auctionSlot, msg.sender, uint128(msg.value));
     }
 
-    function checkOffProposer() external returns (address) {
-        require(msg.sender == rollup, "not a call from the rollup contract");
+    function checkOffProposer() external onlyOwner returns (address) {
         uint32 _currentSlot = currentSlot();
         require(
             auction[_currentSlot].initialized,
--- a/contracts/proposers/BurnAuction.sol
+++ b/contracts/proposers/BurnAuction.sol
@@ -1,5 +1,6 @@
 pragma solidity ^0.5.15;
 import { Chooser } from "./Chooser.sol";
+import { Bitmap } from "../libs/Bitmap.sol";
 
 contract BurnAuction {
     uint32 constant BLOCKS_PER_SLOT = 100;
@@ -20,6 +21,10 @@ contract BurnAuction {
 
     // auction is a relation of the best bid for each slot
     mapping(uint32 => Bid) public auction;
+    // slot -> hasProposed
+    mapping(uint256 => uint256) public hasProposed;
+
+    address public rollup;
 
     /**
      * @dev Event called when an operator beat the bestBid of the ongoing auction
@@ -34,6 +39,7 @@ contract BurnAuction {
      */
     constructor(address _rollup) public {
         require(_rollup != address(0), "Address 0 inserted");
+        rollup = _rollup;
         genesisBlock = getBlockNumber() + DELTA_BLOCKS_INITIAL_SLOT;
     }
 
@@ -58,11 +64,18 @@ contract BurnAuction {
     }
 
     function checkOffProposer() external returns (address) {
+        require(msg.sender == rollup, "not a call from the rollup contract");
+        uint32 _currentSlot = currentSlot();
         require(
-            auction[currentSlot()].initialized,
+            auction[_currentSlot].initialized,
             "Auction has not been initialized"
         );
-        return auction[currentSlot()].operator;
+        require(
+            !Bitmap.isClaimed(uint256(_currentSlot), hasProposed),
+            "Slot fulfilled"
+        );
+        Bitmap.setClaimed(uint256(_currentSlot), hasProposed);
+        return auction[_currentSlot].operator;
     }
 
     /**
--- a/contracts/proposers/BurnAuction.sol
+++ b/contracts/proposers/BurnAuction.sol
@@ -1,7 +1,7 @@
 pragma solidity ^0.5.15;
 import { Chooser } from "./Chooser.sol";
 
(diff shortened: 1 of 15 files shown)


contracts/rollup/Rollup.sol
```diff
@@ -40,10 +41,7 @@ contract RollupSetup {
     uint256 public invalidBatchMarker;
 
     modifier onlyCoordinator() {
-        POB pobContract = POB(
-            nameRegistry.getContractDetails(ParamManager.proofOfBurn())
-        );
-        require(msg.sender == pobContract.getCoordinator());
+        require(msg.sender == chooser.getProposer());
```
ChihChengLiang, line 44: We now have an interface to query who's the proposer. But we need to mark the slot is proposed so that the proposer can't multi proposing in one slot.
Need to think more on this.

contracts/proposers/Chooser.sol
```diff
@@ -0,0 +1,8 @@
+pragma solidity ^0.5.15;
+
+interface Chooser {
```
ChihChengLiang, line 3: Need suggestions for a better name other than "Chooser" 😅

contracts/proposers/BurnAuction.sol
```diff
@@ -0,0 +1,101 @@
...
+    struct Bid {
+        address payable operator;
+        uint128 amount;
+        bool initialized;
+    }
+
+    // auction is a relation of the best bid for each slot
+    mapping(uint32 => Bid) public auction;
+    // slot -> hasProposed
+    mapping(uint256 => uint256) public hasProposed;
```
ChihChengLiang, line 26: Diff to iden3: We track if a winner has proposed in the Burn auction contract, while the iden3 track in the rollup contract.

contracts/proposers/BurnAuction.sol
```diff
@@ -0,0 +1,101 @@
...
+    function bid() external payable {
+        uint32 auctionSlot = currentSlot() + 2;
+        require(
+            msg.value > auction[auctionSlot].amount,
+            "Your bid doesn't beat the current best"
+        );
+        // refund, check 0 case (it means no bids yet for the auction, so no refund)
+        // DISCUSSION: should we refund?
+        if (auction[auctionSlot].initialized && auction[auctionSlot].amount > 0)
+            auction[auctionSlot].operator.transfer(auction[auctionSlot].amount);
```
ChihChengLiang, line 54: Note that solhint warns reentry here.

contracts/proposers/BurnAuction.sol
```diff
@@ -0,0 +1,101 @@
+pragma solidity ^0.5.15;
+import { Chooser } from "./Chooser.sol";
+import { Bitmap } from "../libs/Bitmap.sol";
+import { Ownable } from "@openzeppelin/contracts/ownership/Ownable.sol";
+
+contract BurnAuction is Chooser, Ownable {
```
ChihChengLiang, line 6: Diff to iden3: We let a proposer call rollup contract, which then calls burn auction contract. Iden3 let proposer call burn auction first then rollup contract next.

contracts/proposers/BurnAuction.sol
```diff
@@ -0,0 +1,101 @@
...
+    uint32 constant BLOCKS_PER_SLOT = 100;
+    uint32 constant DELTA_BLOCKS_INITIAL_SLOT = 1000;
```
ChihChengLiang, line 8: We now have 2 new parameters here. And would 100 blocks per batch makes our tps 100x less?

vaibhavchellani: I dont think so, am I missing something? Can a operator push 1 batch per slot only?

contracts/proposers/BurnAuction.sol
```diff
@@ -0,0 +1,101 @@
...
+
+    // First block where the first slot begins
+    uint256 public genesisBlock;
+
+    // Bid is a structure indicating that
+    // "operator" is willing to burn certain
+    // "amount" of ether in order to forge a slot.
+    // "initialized" is used to indicate that the bid is valid (used to differenciate an empty bid)
+    struct Bid {
+        address payable operator;
```
vaibhavchellani, line 18: Can we change this to `coordinator` as `operator` gives the wrong impression?

contracts/proposers/BurnAuction.sol
```diff
@@ -0,0 +1,101 @@
...
+     * @dev Receive a bid from an operator. If the bid is higher than the current bid it replace the existing bid
+     */
+    function bid() external payable {
+        uint32 auctionSlot = currentSlot() + 2;
+        require(
+            msg.value > auction[auctionSlot].amount,
+            "Your bid doesn't beat the current best"
+        );
+        // refund, check 0 case (it means no bids yet for the auction, so no refund)
+        // DISCUSSION: should we refund?
```
vaibhavchellani, line 52: Do you think we should remove the discussion comment here or leave it?

contracts/proposers/BurnAuction.sol
vaibhavchellani: We should probably add a [reentrancy guard](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/ReentrancyGuard.sol) here then

vaibhavchellani: We can also skip this for now and do it for the whole codebase in a single PR if it feels better to go that route.

contracts/proposers/BurnAuction.sol
vaibhavchellani: Curious to know why is it needed to track if a winner hasProposed or not.

contracts/proposers/BurnAuction.sol
```diff
@@ -0,0 +1,101 @@
...
+        if (auction[auctionSlot].initialized && auction[auctionSlot].amount > 0)
+            auction[auctionSlot].operator.transfer(auction[auctionSlot].amount);
+        // set new best bider
+        auction[auctionSlot].operator = msg.sender;
+        auction[auctionSlot].amount = uint128(msg.value);
+        auction[auctionSlot].initialized = true;
+        emit newBestBid(auctionSlot, msg.sender, uint128(msg.value));
+    }
+
+    function checkOffProposer() external onlyOwner returns (address) {
```
vaibhavchellani, line 62: Do you think we need a `view` only getCoordinator?

scripts/deploy.ts
```diff
@@ -29,6 +29,7 @@ async function main() {
         BLOCKS_TO_FINALISE: 7 * 24 * 60 * 4, // 7 days of blocks
         MIN_GAS_LEFT: 10000,
         MAX_TXS_PER_COMMIT: 32,
+        USE_BURN_AUCTION: true,
```
vaibhavchellani, line 32: nice!

vaibhavchellani (approved):
Amazing work! Left a couple of Q's and some nitpicks.

contracts/proposers/BurnAuction.sol
vaibhavchellani: Probably related to another of my comments here https://github.com/thehubbleproject/hubble-contracts/pull/337#discussion_r531596751

contracts/proposers/BurnAuction.sol
ChihChengLiang: If we only allow a winner to propose 1 batch per slot then we need this check.

contracts/proposers/BurnAuction.sol
ChihChengLiang: Skip sounds good. Created an issue for this #352.

contracts/proposers/BurnAuction.sol
ChihChengLiang: Removing it sounds good

contracts/proposers/BurnAuction.sol
ChihChengLiang: If we want to track "hasProposed", then we can't do `view` here.

contracts/proposers/BurnAuction.sol
ChihChengLiang: [winner can propose as many batches they want] so no need to track hasProposed now

contracts/proposers/BurnAuction.sol
ChihChengLiang: [winner can propose as many batches they want] so no harms to tps

contracts/proposers/BurnAuction.sol
ChihChengLiang: [winner can propose as many batches they want] so we can do view here now

PR #339, opened 2020-11-20
Rollback test

ChihChengLiang:


contracts/DepositManager.sol
```diff
@@ -85,17 +91,14 @@ contract DepositCore is SubtreeQueue {
...
+contract DepositManager is DepositCore, IDepositManager, Ownable {
     using Types for Types.UserState;
     Registry public nameRegistry;
     address public vault;
 
     Logger public logger;
     ITokenRegistry public tokenRegistry;
 
-    // batchID => subtreeRoot
-    mapping(uint256 => bytes32) public submittedSubtree;
```
ChihChengLiang, line 97: It makes more sense to let the Rollup contract keep track of which batch has deposit subtree root submitted. So that the deposit contract can be batch agnostic.

contracts/rollup/BatchManager.sol
```diff
@@ -0,0 +1,156 @@
...
+            batchID--
+        ) {
+            if (gasleft() <= paramMinGasLeft) break;
+
+            Bitmap.setClaimed(batchID, withdrawalBitmap);
+            delete batches[batchID];
+
+            depositSubTreeRoot = deposits[batchID];
+            if (depositSubTreeRoot != bytes32(0))
+                depositManager.reenqueue(depositSubTreeRoot);
```
ChihChengLiang, line 90: We now check if a batch is a deposit batch before doing the external call.

contracts/rollup/Rollup.sol
```diff
@@ -546,31 +350,58 @@ contract Rollup is RollupHelpers {
...
 
-    /**
-     * @notice Withdraw delay allows coordinators to withdraw their stake after the batch has been finalised
-     */
-    function withdrawStake(uint256 batchID) public {
-        require(
-            msg.sender == batches[batchID].committer(),
-            "You are not the correct committer for this batch"
+contract Rollup is RollupCore {
+    constructor(
```
ChihChengLiang, line 358: Move the constructor to the tail of the inheritance chain, so that testing is easier.

contracts/test/TestRollup.sol
```diff
@@ -0,0 +1,54 @@
...
+import { IDepositManager } from "../DepositManager.sol";
+
+contract MockDepositManager is IDepositManager {
+    event EnqueSubtree(bytes32 subtreeRoot);
+
+    function dequeueToSubmit() external returns (bytes32 subtreeRoot) {
+        return bytes32(0);
+    }
+
+    function reenqueue(bytes32 subtreeRoot) external {
```
ChihChengLiang, line 16: This biased the gas profiling of rollback because the mocked reenqueue is not as expensive as the real depositManager.

vaibhavchellani (approved):
Amazing work @ChihChengLiang! Do we have some benchmarks yet?

ChihChengLiang:
Thank you for the review! And yes we have some benchmark. 
it takes 31k when the loop exits, and on average it takes 10k gas to rollback a batch. The numbers are logged in the CI so we can watch the changes in every pr.

PR #343, opened 2020-11-24
immutable paramMaxSubtreeSize

ChihChengLiang:
Since we have decided to make parameters immutable, we can simplify the setup in the deposit manager.


vaibhavchellani (approved):
Looks perfect!

PR #346, opened 2020-11-25
Make batches a mapping

### LINKED ISSUE #338: Make Batch a mapping
We currently define batch as an array in storage, this comes with some disadvantages:

- Need to keep track of `batches.length`
- Can't simply delete a finalized batch

ChihChengLiang:
Fix #338

### DIFF
--- a/contracts/rollup/BatchManager.sol
+++ b/contracts/rollup/BatchManager.sol
@@ -4,6 +4,7 @@ pragma experimental ABIEncoderV2;
 import { SafeMath } from "@openzeppelin/contracts/math/SafeMath.sol";
 import { Types } from "../libs/Types.sol";
 import { Parameters } from "./Parameters.sol";
+import { Bitmap } from "../libs/Bitmap.sol";
 import { IDepositManager } from "../DepositManager.sol";
 
 contract BatchManager is Parameters {
@@ -20,6 +21,9 @@ contract BatchManager is Parameters {
 
     // batchID -> depositSubtreeRoot
     mapping(uint256 => bytes32) public deposits;
+
+    // batchID -> hasWithdrawn
+    mapping(uint256 => uint256) public withdrawalBitmap;
 
     // this variable will be greater than 0 if
     // there is rollback in progress
@@ -129,9 +133,6 @@ contract BatchManager is Parameters {
      * @notice Withdraw delay allows coordinators to withdraw their stake after the batch has been finalised
      */
     function withdrawStake(uint256 batchID) public {
-        // A valid batch should have non-zero meta
-        // We use this to check if a stake has been withdrawn
-        require(batches[batchID].meta != bytes32(0), "Rollup: No such batch");
         require(
             msg.sender == batches[batchID].committer(),
             "You are not the correct committer for this batch"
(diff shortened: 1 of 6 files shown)


contracts/rollup/BatchManager.sol
```diff
@@ -143,11 +140,8 @@ contract BatchManager is Parameters {
             block.number > batches[batchID].finaliseOn(),
             "This batch is not yet finalised, check back soon!"
         );
-        require(
-            !Bitmap.isClaimed(batchID, withdrawalBitmap),
-            "Rollup: Already withdrawn"
-        );
-        Bitmap.setClaimed(batchID, withdrawalBitmap);
+        // delete the batch to mark the stake has been withdrawn
+        delete batches[batchID];
```
ChihChengLiang, line 144: We now track the withdrawal with the existence of the batch.
📈 simplicity/ gas saving on storage
📉 the logic is indirect

contracts/rollup/BatchManager.sol
```diff
@@ -127,14 +121,17 @@ contract BatchManager is Parameters {
...
+        nextBatchID++;
     }
 
     /**
      * @notice Withdraw delay allows coordinators to withdraw their stake after the batch has been finalised
      */
     function withdrawStake(uint256 batchID) public {
+        // A valid batch should have non-zero meta
+        // We use this to check if a stake has been withdrawn
+        require(batches[batchID].meta != bytes32(0), "Rollup: No such batch");
```
ChihChengLiang, line 134: The meta has fields like batchType, commitmentLength, committer, finaliseOn. The last two should guarantee the existent batch have a non-zero meta.

contracts/rollup/BatchManager.sol
```diff
@@ -127,14 +121,17 @@ contract BatchManager is Parameters {
                 block.number + paramBlocksToFinalise
             )
         });
-        batches.push(newBatch);
-        emit NewBatch(msg.sender, batches.length - 1, batchType);
+        emit NewBatch(msg.sender, nextBatchID, batchType);
+        nextBatchID++;
```
ChihChengLiang, line 125: A downside is we have to carefully track `nextBatchID`

contracts/rollup/BatchManager.sol
ChihChengLiang: Ah! Bad idea deleting batches. The vault still needs it.

contracts/rollup/BatchManager.sol
```diff
@@ -79,25 +78,24 @@ contract BatchManager is Parameters {
...
         uint256 totalSlashings = 0;
         for (
-            uint256 batchID = batches.length - 1;
+            uint256 batchID = nextBatchID - 1;
             batchID >= invalidBatchMarker;
             batchID--
         ) {
             if (gasleft() <= paramMinGasLeft) break;
 
-            Bitmap.setClaimed(batchID, withdrawalBitmap);
```
ChihChengLiang, line 88: We can't `setClaimed` here.
- The malicious committer can't claim the stake because the batch is deleted.
- If setClaimed, a future good committer will not be able to claim the batch of this batchID.

vaibhavchellani: Good catch

contracts/rollup/BatchManager.sol
vaibhavchellani: Do you want to create some internal methods like `add` that would take care of this automatically?

vaibhavchellani (approved):
LGTM!

contracts/rollup/BatchManager.sol
ChihChengLiang: We have 2 places that use nextBatchID++, should be manageable without the internal method for now.

PR #355, opened 2020-11-29
make batch account registration size smaller

### LINKED ISSUE #350: Allow dynamic pubkey batch size
Would be great if we could allow dynamic size of pubkeys while [registering](https://github.com/thehubbleproject/hubble-contracts/blob/master/contracts/BLSAccountRegistry.sol#L16) instead of the current 1024.

ChihChengLiang:
Fix #350. Choose a smaller power of 2.


vaibhavchellani (approved):
perfect!

PR #359, opened 2020-12-01
Better genesis and Integration tests

ChihChengLiang:


ChihChengLiang:
Close in favor of #386

PR #361, opened 2020-12-01
Fix create2 signed message 

### LINKED ISSUE #360: Remove from pubkey from messageOf
We earlier required the fromPubkey to be signed however that leads to more DB queries so we can just replace that with fromIndex.

vaibhavchellani:
Fixes #360


ChihChengLiang (approved):
Looks good to me!

PR #364, opened 2020-12-03
fix account tree log

ChihChengLiang:


ChihChengLiang:
The original one is the intended behavior. We have the left tree for a single update and the right tree for the batch update. that's why the log emits the index that's the "whole set size of the left tree" + "existing right tree index" + "current index in the batch".

PR #366, opened 2020-12-03
Revamp state tree again

ChihChengLiang:
### What's wrong

- In one situation we want the state tree to append placeholder states and witnesses when a state transition fails so that we can simulate a fraudulent commitment.
- In other situations, we want to debug why a certain state transition fails.

### How can we fix that

`stateTree.processTransferCommit(..., raiseError: true)`

We add a raiseError arg in the API. If the arg is set, then the function raises errors, otherwise, it silently appends dummy proofs.

This is an example error message I'm seeing in my other local branch.
```
1) Integration Test
       Getting new users via Create to transfer:
     Error: balance: 20000000, tx amount+fee: 110000000
      at StateTree.processSender (ts/stateTree.ts:274:19)
      at StateTree.processCreate2Transfer (ts/stateTree.ts:245:34)
      at StateTree._processCreate2TransferCommit (ts/stateTree.ts:159:55)
      at _processCreate2TransferCommit.next (<anonymous>)
      at Function.from (<anonymous>)
      at StateTree.processCreate2TransferCommit (ts/stateTree.ts:187:36)
      at Context.<anonymous> (test/integration.test.ts:219:40)
```
We achieve the functionality by using the Generator. Say, we have 10 Transfer txs, we expect to have 21 proofs (10 sender+ 10 receiver + fee). So if the 5th tx is a bad state transition and an error is raised, the generator can still successfully yield valid proofs for the 4 good txs.

### DIFF
--- a/test/create2transfer.test.ts
+++ b/test/create2transfer.test.ts
@@ -169,9 +169,10 @@ describe("Rollup Create2Transfer Commitment", () => {
 
         for (const tx of txs) {
             const preRoot = stateTree.root;
-            const { proofs, safe } = stateTree.processCreate2Transfer(tx);
-            const [senderProof, receiverProof] = proofs;
-            assert.isTrue(safe);
+            const [
+                senderProof,
+                receiverProof
+            ] = stateTree.processCreate2Transfer(tx);
             const postRoot = stateTree.root;
             const {
                 0: processedRoot,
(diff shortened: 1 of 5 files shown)


ChihChengLiang:
Need this change now. Post merging review is welcomed.

PR #367, opened 2020-12-04
Create2

ChihChengLiang:


ts/deploy.ts
```diff
@@ -20,32 +17,37 @@ import {
...
+    public signer: Signer;
+    constructor(public deployerContract: Deployer) {
+        this.signer = this.deployerContract.signer;
+    }
+    public async deploy(
+        request: providers.TransactionRequest
+    ): Promise<string> {
+        const signerAddress = await this.signer.getAddress();
+        const salt = id(signerAddress);
+        const data = request.data as string;
```
ChihChengLiang, line 42: We are not able to resolve the circular dependency(two contracts' constructor depends on other's address) with CREATE2, because the CREATE2 address depends on the constructor args

ChihChengLiang:
Closing. We'll work on other deployment strategies

PR #368, opened 2020-12-04
Better genesis 2

ChihChengLiang:
### Behavior change

1. We don't create genesis root in the ts/deploy.ts now. Genesis must be explicitly specified in the deployment parameters, otherwise ts/deploy.ts raises an error.
2. The scripts/deploy.ts has default genesis root if the genesis root is not specified in the arg. So client testing can be like the good old days. Note that the new genesis states have a large amount of balance.

### DIFF
--- a/scripts/deploy.ts
+++ b/scripts/deploy.ts
@@ -6,6 +6,8 @@ import { toWei } from "../ts/utils";
 import { BlsAccountRegistry } from "../types/ethers-contracts/BlsAccountRegistry";
 import fs from "fs";
 import { PRODUCTION_PARAMS } from "../ts/constants";
+import { StateTree } from "../ts/stateTree";
+import { State } from "../ts/state";
 
 const argv = require("minimist")(process.argv.slice(2), {
     string: ["url", "root", "pubkeys"]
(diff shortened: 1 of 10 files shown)


ts/deploy.ts
```diff
@@ -256,18 +253,3 @@ export async function deployAll(
...
-    parameters: DeploymentParameters
-) {
-    const state0 = State.new(0, 0, 0, 0);
-    const state1 = State.new(1, 0, 0, 0);
-    const dataLeaves = [state0.toStateLeaf(), state1.toStateLeaf()];
-    const ZERO_BYTES32 =
-        "0x290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563";
-    for (let i = dataLeaves.length; i < 2 ** parameters.MAX_DEPTH; i++) {
-        dataLeaves[i] = ZERO_BYTES32;
-    }
```
ChihChengLiang, line 270: Building the 32 levels of the leaves is why the deploy script freezes a second before deploying the rollup contract.

test/MassMigration.test.ts
```diff
@@ -14,6 +14,7 @@ import { Result } from "../ts/interfaces";
 import { Tree } from "../ts/tree";
 import { mineBlocks } from "../ts/utils";
 import { expectRevert } from "../ts/utils";
+import { constants } from "ethers";
 
 describe("Mass Migrations", async function() {
```
ChihChengLiang, line 19: I intend to replace the whole `test/rollup.test.ts`, `test/MassMigration.test.ts`, and `test/rollupCreate2Transfer.test.ts` with the new integration test. So I only attempt to fix the first test, but bandaid the latter two tests with dummy genesis roots.

scripts/deploy.ts
```diff
@@ -17,6 +19,17 @@ const argv = require("minimist")(process.argv.slice(2), {
     --pubkeys 06642b1af0ec2f7369126b3e45aaf11a050a26e2111150b319c4ca2f4a8d48c62442e01b651ce469632972f06c4d5d092da2cde3d42c0ead5ebeed8646d72b3f1f09c41c70cbdd8779920b4a3b15e6c7c518c5fdb93fdaed7a49c59b1f4d1e210aafa7a17239c9df9053ca4e71f10fb9f7b0f219e12e300676a756463253c287,2a7575b1bddf2c2b25f976788baae059ea410f589f0c244e1554d6f6f5398b6a1b997670d19c26c33ad1c61e21bd2742565518e0deafb105e7718de983bf757a0770cb4889aabb6f4897534f4770fce4b76f39b2cc8ed63b22e55ebf7e87a96001bbabd11ca55585bc0c016ad058a69c682cbe102a4b91f6084c71c857509dac
 */
 
+function getDefaultGenesisRoot(parameters: DeploymentParameters) {
+    const stateTree = StateTree.new(parameters.MAX_DEPTH);
+    const LARGE_AMOUNT = 1000000000;
+    const state0 = State.new(0, 0, LARGE_AMOUNT, 0);
```
vaibhavchellani, line 25: why do we initialise with LARGE_AMOUNT instead of `0`?

vaibhavchellani (approved):
LGTM!

scripts/deploy.ts
ChihChengLiang: So it might be the case that the coordinator wants to spend some token ASAP. Note that we now don't require special states for Mass migration or create2transfer. The default genesis states are created just for convenience.

vaibhavchellani: This will make the `leaf_hash != defaultHashes(0)` which will lead to inconsistencies on the client. If this change is not required by some feature then think we should skip it.

ChihChengLiang: Sounds good to me.

PR #372, opened 2020-12-07
Revamp mcl and add BLS signer abstraction

ChihChengLiang:
### What's wrong

Writing the integration test is difficult now because our State class handles too many things at once. Ideally, the state should have these properties pubkeyID, tokenID, balance, and nonce, and manipulate those properties with utilities like toLeaf, encode.

Instead, the State class also mimics the user, holds private keys, and signs stuff. This creates a problem that we need to clone or copy the keys when doing create to transfer.

### How are we fixing that

This is the first step of the fix. We first create a blsSigner class that handles the key generation, holds the key, and signs messages. It has more benefits:
- It holds domain so we no longer need to set the global one.
- no need to manipulate mcl logic. blsSigner translates values to Solidity structs.

We'll separate signer and state in the next PR.


ts/blsSigner.ts
```diff
@@ -0,0 +1,60 @@
...
+    aggregateRaw,
+    mclG1,
+    solG1,
+    SecretKey,
+    randFr,
+    PublicKey
+} from "./mcl";
+
+export interface SignatureInterface {
+    mcl: mclG1;
```
ChihChengLiang, line 17: keeping mclG1 here so that `aggregate` can work on it directly without converting solG1 back to mclG1.

vaibhavchellani (approved):
LGTM!

PR #373, opened 2020-12-07
Rename TestToken to ExampleToken

ChihChengLiang:
Just don't want this in the audit scope. 

When time permits we might want to remove this contract from the deployment script.


vaibhavchellani (approved):
LGTM

PR #375, opened 2020-12-08
Fix frontend contract: Mass Migration and Create2Transfer

### LINKED ISSUE #370: Create2 frontend invalid token check
The create2 [validate and apply frontend function](https://github.com/thehubbleproject/hubble-contracts/blob/master/contracts/libs/Transition.sol#L204) does a token type check on receiver which fails because receiver doesn't exist while doing create2.

ChihChengLiang:
Fix #370. Mass Migration and Create2Transfer have no receiver


vaibhavchellani (approved):
LGTM

PR #376, opened 2020-12-08
Rename zero/empty state

ChihChengLiang:
### What's wrong

Two situations of state leaf could be confusing:

- `keccak(abi.encode(state.pubkeyID, state.tokenID, state.balance, state.nonce))` is `keccak(abi.encode(0, 0, 0, 0))`
- `keccak(abi.encode(0))`

I propose we call the former a zero leaf, and the latter an empty leaf or a null leaf. The reason is if we treat the properties of the state as a vector of values, the first case is like a zero vector and the second case is like an empty set.

I believe we used to call the second case a zero state because we call that hash a zero_hash. But we could replace that hash with other Nothing-up-my-sleeve numbers.


vaibhavchellani (approved):
LGTM

PR #379, opened 2020-12-08
Revamp state tree 3

ChihChengLiang:
- More abstraction on the stateTree. At the core, all operation is just `getState` or `updateState`, or the two combined.
- Remove StateSolStruct, the State class itself is enough to work with the contract
- Handles the tokenID validation correctly, they mirror the contract now.


kilic (commented):
Looks good to me.

PR #384, opened 2020-12-10
Moving the signing methods from State to User

ChihChengLiang:
### What's wrong

The current way that State signs stuff and holds balance lets us shoot ourselves in the foot.

- Key is lost when a state is created or cloned
- Balance is outdated
- Mock state has the source of truth of keys and the state tree has the source of truth of balance

### How we are fixing it

We create a User class that knows only their stateID and pubkeyID and they sign stuff.

State only holds balances and the state tree has the single source of truth of balances.

We also created a Group class to handle a group of users gracefully.

### DIFF
--- a/ts/factory.ts
+++ b/ts/factory.ts
@@ -28,6 +28,9 @@ export class User {
     get pubkey() {
         return this.blsSigner.pubkey;
     }
+    toString() {
+        return `<User stateID: ${this.stateID}  pubkeyID: ${this.pubkeyID}>`;
+    }
 }
 
 interface GroupOptions {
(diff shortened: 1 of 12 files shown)


ts/factory.ts
```diff
@@ -1,114 +1,224 @@
...
+    initialPubkeyID?: number;
+}
+
+interface createStateOptions {
+    initialBalance?: BigNumber;
+    tokenID?: number;
+    zeroNonce?: boolean;
+}
+
+export class Group {
```
vaibhavchellani, line 63: very nice

vaibhavchellani (approved):
Solid work!

ChihChengLiang:
Thank you for the review 😄

PR #385, opened 2020-12-11
Burn auction upgrade

kilic:


test/burnAuction.test.ts
```diff
@@ -1,18 +1,28 @@
...
 import { ethers } from "hardhat";
 import { toWei } from "../ts/utils";
 import { MockRollup } from "../types/ethers-contracts/MockRollup";
 import { MockRollupFactory } from "../types/ethers-contracts/MockRollupFactory";
 import { TestBurnAuction } from "../types/ethers-contracts/TestBurnAuction";
 import { TestBurnAuctionFactory } from "../types/ethers-contracts/TestBurnAuctionFactory";
 
-const defaultAddress = "0x0000000000000000000000000000000000000000";
+const {
+    expectRevert // Assertions for transactions that should fail
```
ChihChengLiang, line 13: FYI we have expectRevert in the `./ts/utils.ts` too, don't need to add @openzeppelin/test-helpers dependency if that works.

contracts/proposers/BurnAuction.sol
```diff
@@ -66,6 +101,26 @@ contract BurnAuction is Chooser {
...
+    }
+
+    function witdraw(uint256 amount) external {
+        address payable claimer = msg.sender;
+        require(
+            deposits[claimer] >= amount,
+            "BurnAuction, withdraw: insufficient deposit amount for withdraw"
+        );
+        claimer.transfer(amount);
+        updateBalance(claimer, 0, amount);
```
ChihChengLiang, line 121: We should update before a transfer to prevent a reentrance.
```suggestion
        updateBalance(claimer, 0, amount);
        claimer.transfer(amount);
```

contracts/proposers/BurnAuction.sol
```diff
@@ -22,39 +32,64 @@ contract BurnAuction is Chooser {
...
+            bidAmount.mul(donationNumerator).div(DONATION_DENOMINATOR),
+            currentBidAmount.mul(donationNumerator).div(DONATION_DENOMINATOR)
         );
-        // refund, check 0 case (it means no bids yet for the auction, so no refund)
-        if (auction[auctionSlot].initialized && auction[auctionSlot].amount > 0)
-            auction[auctionSlot].coordinator.transfer(
-                auction[auctionSlot].amount
-            );
+        // update coordinator with remaining value
+        updateBalance(coordinator, msg.value, bidAmount);
```
ChihChengLiang, line 87: So is it correct that the coordinator:
1. specifies the `bidAmount` to bid
2. and choose to
- pay the bid with `msg.value`
- pay the bid with remaining deposits
- or pay the bid with `msg.value` and deposits combined
- or purely bid to deposit values for future use.

ts/constants.ts
```diff
@@ -18,7 +20,9 @@ export const PRODUCTION_PARAMS: DeploymentParameters = {
     BLOCKS_TO_FINALISE: 7 * 24 * 60 * 4, // 7 days of blocks
     MIN_GAS_LEFT: 10000,
     MAX_TXS_PER_COMMIT: 32,
-    USE_BURN_AUCTION: true
+    USE_BURN_AUCTION: true,
+    DONATION_ADDRESS: "0x00000000000000000000000000000000000000d0",
+    DONATION_NUMERATOR: 7500
```
ChihChengLiang, line 25: I think it is fine we hardcode the donation address and donation numerator/percentage in the deployment config. It's beneficial to leave them in the commit history.
We could update in the future when we have a clear target to donate.

ChihChengLiang (approved):
Perfect work! 
I found a reentrance issue that needs to address.
Other parts look great to me.

contracts/proposers/BurnAuction.sol
kilic: Somehow easier reasoning can go like this:
1. msg.value adds up to the deposit.
2. bidAmount is paid from deposit.
3. remaining value stays in the deposit.

ChihChengLiang: Oh, that sounds easier to understand.

PR #388, opened 2020-12-11
Fix batch update

ChihChengLiang:
### What's wrong

The leafIndexRight should increase by a batch size on every batch update, but it currently only increase by 1. That's why the `PubkeyRegistered` starts to emit the incorrect pubkeyIDs since the second batch update.

### DIFF
--- a/contracts/AccountTree.sol
+++ b/contracts/AccountTree.sol
@@ -82,7 +82,7 @@ contract AccountTree {
         bytes32 leaf = leafs[0];
 
         // Ascend to the root
-        uint256 path = leafIndexRight;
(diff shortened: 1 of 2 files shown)


ChihChengLiang (commented):
@vaibhavchellani This might be the bug that's troubling you

vaibhavchellani:
@ChihChengLiang ah! Yes it is, thanks!

kilic (commented):
Looks great!

ChihChengLiang:
@vaibhavchellani Np :) Glad to know it helps.
@kilic Thanks for the review!

PR #386, opened 2020-12-11
Integration test 2

ChihChengLiang:
Should be merged after #389

Close #221

PR #389, opened 2020-12-12
Revamp trees

ChihChengLiang:
### What's wrong

- AccountTree didn't handle 32 depths well
- Tree fails silently

### How are we fixing it

Instead of wrangling everything in BigNumbers or doing JS integers correctly, we raise explicit exceptions and add failing inputs to the error message stack to speed up dev iteration.

### DIFF
--- a/ts/accountTree.ts
+++ b/ts/accountTree.ts
@@ -1,15 +1,17 @@
 import { Tree } from "./tree";
 import { BlsAccountRegistry } from "../types/ethers-contracts/BlsAccountRegistry";
 import { ethers } from "ethers";
-import { assert } from "chai";
+import { solG2 } from "./mcl";
+import { RegistrationFail } from "./exceptions";
 
+// Tree is 32 level depth, the index is still smaller than Number.MAX_SAFE_INTEGER
 export class AccountRegistry {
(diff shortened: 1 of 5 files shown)


ts/accountTree.ts
```diff
@@ -25,34 +27,78 @@ export class AccountRegistry {
     ) {
         this.treeLeft = Tree.new(depth);
         this.treeRight = Tree.new(depth);
-        this.setSize = 1 << depth;
+        this.setSize = 2 ** depth;
```
ChihChengLiang, line 30: JS bitwise operation is signed. Power operation is a way to get correct 31 depth value
Number.MAX_SAFE_INTEGER is `2**53 -1`, so we are still safe in Number.

```
$ node
Welcome to Node.js v12.14.1.
Type ".help" for more information.
> 1 << 31
-2147483648
> 2 ** 31
2147483648
> (1 << 31) >>> 0
2147483648
```

ChihChengLiang:
The change in this PR is mostly not tested. But they are used and covered in #386.

vaibhavchellani:
Nice find!

kilic (approved):
Looks good to me.

PR #393, opened 2020-12-18
Long running tests

ChihChengLiang:
- Fail fast: Run fast tests and unit tests first so that failure can be detected as early as possible.
- Rename commitment/rollup tests properly. The previous naming confuses the following tests
  - commit.* tests are those wrapped with test contracts, and
  - rollup.* tests are those tested against the rollup contract.


vaibhavchellani (approved):
Looks good to me

PR #395, opened 2020-12-24
Added deposit script

vaibhavchellani:
Fixes #394

PR #396, opened 2020-12-25
Add Hubble class and hubble repl

ChihChengLiang:
### What's wrong

When we want to do something with the deployed contracts. We need to write a lengthy script like #395.

### How are we fixing it

We can create a Hubble class that parses the genesis config and manages all contract instances. So whenever we want to do something with any contract, we access them from the Hubble class. 

We can also extend convenient methods to do insert pubkeys or submit deposits smoothly.

What's best is we can do repl with it. You can run `npm run repl` then it prompts you.

To reproduce this example, we need to run `npm run node` in another terminal and run `npm run deploy`.

The repl instance expose `hubble` and `ethers` instances in the context. So we can access the contract and manipulate the data types with those.

```
hubble > hubble.contracts.exampleToken.approve(hubble.contracts.depositManager.address, ethers.utils.parseEther("100"))
Promise { <pending> }
hubble > hubble.contracts.depositManager.depositFor(0, 100, 0)
Promise { <pending> }
```

### DIFF
--- a/scripts/deploy.ts
+++ b/scripts/deploy.ts
@@ -2,19 +2,17 @@ import { ethers } from "ethers";
 import { allContracts } from "../ts/allContractsInterfaces";
 import { deployAll } from "../ts/deploy";
 import { DeploymentParameters } from "../ts/interfaces";
-import { BlsAccountRegistry } from "../types/ethers-contracts/BlsAccountRegistry";
 import fs from "fs";
 import { PRODUCTION_PARAMS } from "../ts/constants";
 import { StateTree } from "../ts/stateTree";
 
 const argv = require("minimist")(process.argv.slice(2), {
-    string: ["url", "root", "pubkeys"]
+    string: ["url", "root"]
 });
 /*
     Note separate pubkeys with commas
     > npm run deploy -- --url http://localhost:8545 \
-    --root 0x309976060df37ed6961ebd53027fe0c45d3cbbbdfc30a5039e86b2a7aa7fed6e \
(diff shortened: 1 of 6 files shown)


scripts/deploy.ts
```diff
@@ -44,30 +42,6 @@ async function main() {
...
     fs.writeFileSync("genesis.json", JSON.stringify(configs, null, 4));
     console.log("Successsfully deployed", configs);
-
-    if (argv.pubkeys) {
-        const pubkeys = argv.pubkeys.split(",");
-        await registerPublicKeys(contracts.blsAccountRegistry, pubkeys);
-    }
-}
-
-async function registerPublicKeys(
```
ChihChengLiang, line 54: Remove the public key registration from the deployment stage. Since we want to do it with deposits

scripts/deposit.ts
```diff
@@ -1,96 +1,45 @@
...
     let accIDs: number[] = [];
 
     if (argv.pubkeys) {
         const pubkeys = argv.pubkeys.split(",");
-        accIDs = await registerPublicKeys(accountRegistry, pubkeys);
+        accIDs = await hubble.registerPublicKeys(pubkeys);
     }
+    const { tokenRegistry, exampleToken, depositManager } = hubble.contracts;
 
+    const tokenID = 0;
```
ChihChengLiang, line 21: Since we are checking against the exampleToken, we can remove tokenID from the cli args and hardcode it here.

scripts/deposit.ts
```diff
@@ -1,96 +1,45 @@
...
 
     // approve depositmanager for amount
-    let amount = toWei(argv.amount);
+    const amount = parseInt(argv.amount);
+    const totalAmount = accIDs.length * amount;
     let approveTx = await exampleToken.approve(
-        genesis.addresses["depositManager"],
-        toWei(argv.amount)
+        depositManager.address,
+        toWei(totalAmount.toString())
```
ChihChengLiang, line 34: To deposit 5 tokens for each 5 public keys, we need to approve 25 tokens here.

scripts/deposit.ts
vaibhavchellani: Would like to have the option of using this script for not only `exampleToken` but for other tokens too.

scripts/deposit.ts
vaibhavchellani: Yeah, I fixed this locally and forgot to push, thanks!

vaibhavchellani (approved):
This is very nice, thanks @ChihChengLiang

PR #404, opened 2021-01-07
Fix param manager

### LINKED ISSUE #401: [Audit] Use hash result in paramManager
https://github.com/thehubbleproject/hubble-contracts/blob/32234da0ce3015b4bece04a264a136a6778191dd/contracts/libs/ParamManager.sol#L3-L65
> Solidity does not compute keccak at compile time. We recommend replacing hash execution with final results and adding preimages in comments.

ChihChengLiang:
Fix #401. And also

- Remove unused functions
- ~~make functions internal~~ Can't make the function internal, otherwise we can't register in the name registry.


ChihChengLiang:
Close in favor of #405

Issue #417, opened 2021-01-11
Verification failure when 4 or less signatures are aggregated.


jzaki:
https://github.com/thehubbleproject/hubble-contracts/blob/3c6e39b2b0f1b7969ae4be90fdb4fa0036735644/test/fast/bls.test.ts#L67
I didn't come across a documented minimum.

**Steps**:
- set `n = 4` (or 2, 3)
- run tests

**Result**: Verification fails. **Expect**: Verification to succeed

ChihChengLiang:
Nice catch! Looks like the miscalculation of the gas cost cause it. 
https://github.com/ethereum/go-ethereum/blob/master/params/protocol_params.go#L127

ChihChengLiang:
Sending a fix #418. Thank you @jzaki for finding this issue

ChihChengLiang:
Fixed in #418. Feel free to reopen if you find it doesn't work

PR #418, opened 2021-01-12
fix bls gas

ChihChengLiang:
### What's wrong

#417 BLS check failed when less or equal than 4 messages.

This is due to 

1. we misplaced the base cost and the per pairing cost of the precompile call. See 
https://github.com/ethereum/go-ethereum/blob/master/params/protocol_params.go#L127 for the ground truth.
2. checking n public keys needs n + 1 pairings

### How can we fix it

Fix the incorrect gas cost computation.

### Note

Could be superseded by #409, as we are trying a new approach to calculate the gas cost

### DIFF
--- a/contracts/libs/BLS.sol
+++ b/contracts/libs/BLS.sol
@@ -124,8 +124,8 @@ library BLS {
         pure
         returns (uint256)
     {
-        uint256 base = 34000;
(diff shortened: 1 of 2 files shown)


kilic (approved):
Looks good

PR #428, opened 2021-01-15
remove dead code

ChihChengLiang:
It's been a while we haven't used those and assume we'll never use them again.


vaibhavchellani (approved):
LGTM!

PR #430, opened 2021-01-18
Debug precompile cost estimator

ChihChengLiang:


ChihChengLiang:
debug is done here and concluded in #448

PR #431, opened 2021-01-19
organizing 32 level depth

ChihChengLiang:
- Test integration test with 32 levels deep state tree.
- cleanup references to commit size and tree depth 32


vaibhavchellani (approved):
LGTM

Issue #432, opened 2021-01-19
Use events more efficiently in the Rollback


ChihChengLiang:
### What's wrong

See feedback https://github.com/thehubbleproject/hubble-contracts/pull/429#pullrequestreview-569935841

We currently fire an event per batch rollback, which is inefficient. Instead, we should only fire an event when the rollback finished. 

### How can we fix it

- ~~emit BatchesSlashed(start, stop) at the end of the loop~~
- `emit RollbackCompleted(start, stop)` or `emit PartiallyCompleted(start, stop)` depending on the result

ChihChengLiang:
Fixed in #433

PR #433, opened 2021-01-19
Rollback events

ChihChengLiang:
Address #432 

### Gas cost

This PR

- Rolled back 249 batches with 1234263 gas
- gas per batch rolled back 4956.879518072289

Master

- Rolled back 249 batches with 1375402 gas
- gas per batch rolled back 5523.70281124498

### DIFF
--- a/contracts/rollup/BatchManager.sol
+++ b/contracts/rollup/BatchManager.sol
@@ -83,7 +83,7 @@ contract BatchManager is Parameters {
         uint256 nExpect = nextBatchID - minID;
         uint256 nActual = 0;
         while (nActual < nExpect) {
-            batchID = minID + nExpect - nActual;
+            batchID = minID + nExpect - 1 - nActual;
             if (gasleft() <= paramMinGasLeft) break;
 
             delete batches[batchID];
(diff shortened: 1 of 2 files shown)


contracts/rollup/BatchManager.sol
```diff
@@ -34,8 +34,8 @@ contract BatchManager is Parameters {
 
     event NewBatch(address committer, uint256 index, Types.Usage batchType);
     event StakeWithdraw(address committed, uint256 batchID);
-    event BatchRollback(uint256 batchID);
-    event RollbackFinalisation(uint256 totalBatchesSlashed);
+    event RollbackComplete(uint256 startID, uint256 nDeleted);
+    event RollbackPartiallyComplete(uint256 startID, uint256 nDeleted);
```
ChihChengLiang, line 38: I found that using endID at the second parameter might be confusing.
- If endID represents the ID of the last batch deleted, then it's not possible to represent the case that no batch is deleted.
- It could represent the ID of the highest batch that hasn't been deleted. But a little bit harder to reason than simplly the number of how many batches deleted.

vaibhavchellani (approved):
LGTM

kilic (approved):

PR #434, opened 2021-01-19
add version field

ChihChengLiang:
so that we know what contract versions to track for a specific deployment.


vaibhavchellani (approved):
LGTM

PR #435, opened 2021-01-21
Fix public network deployments

sz-piotr:
This PR fixes a bug in which the deployment fails on a public network like ropsten, because the transactions were not yet mined. It also adds an option to specify the private key from the command line.

### DIFF
--- a/scripts/deploy.ts
+++ b/scripts/deploy.ts
@@ -8,12 +8,15 @@ import { StateTree } from "../ts/stateTree";
 import { execSync } from "child_process";
 
 const argv = require("minimist")(process.argv.slice(2), {
-    string: ["url", "root"]
+    string: ["url", "root", "key"]
 });
 /*
     Note separate pubkeys with commas
     > npm run deploy -- --url http://localhost:8545 \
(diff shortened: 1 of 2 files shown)


scripts/deploy.ts
```diff
@@ -24,11 +27,12 @@ function getDefaultGenesisRoot(parameters: DeploymentParameters) {
...
-    const signer = provider.getSigner();
-    // const signer = new ethers.Wallet(
-    //     "0xyourprivatekeyhere"
-    // ).connect(provider);
+    const provider = new ethers.providers.JsonRpcProvider(
+        argv.url ?? "http://localhost:8545"
+    );
+    const signer = argv.key
+        ? new ethers.Wallet(argv.key).connect(provider)
+        : provider.getSigner();
```
ChihChengLiang, line 35: Beautiful! With this change, we can gracefully pass the private key to the command argument. I like it so much.

ChihChengLiang (approved):
Nice work! I tested locally and it deploys successfully. https://goerli.etherscan.io/address/0x6cd0bb5e580781942b2216594da67cc80a93c539

PR #445, opened 2021-01-22
changed production params

vaibhavchellani:


vaibhavchellani:
Let's not increase the stake amount

Issue #427, opened 2021-01-15
AccountTree: Input data could be damaged.


ChihChengLiang:
> Input data could be damaged.

https://github.com/thehubbleproject/hubble-contracts/blob/32234da0ce3015b4bece04a264a136a6778191dd/contracts/AccountTree.sol#L65

ChihChengLiang:
Fixed in #450

PR #450, opened 2021-01-23
Make batch update function non destructive

kilic:
#427


ChihChengLiang (approved):
Great work! Now batch update doesn't damage the input. Looks good to me.

PR #452, opened 2021-01-28
set account tree first default hash to be 0x290...

ChihChengLiang:
### What's wrong

Currently, we have different default hashes for the state tree and the public key tree. This might confuse client developers and application developers

- first default hash for state tree: `0x290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563`
- first default hash for public key tree: `0x0000000000000000000000000000000000000000000000000000000000000000`

### How are we fixing it

Make them the same. Set to `0x290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563`

### DIFF
--- a/contracts/AccountTree.sol
+++ b/contracts/AccountTree.sol
@@ -25,6 +25,13 @@ contract AccountTree {
     bytes32[DEPTH - BATCH_DEPTH] public filledSubtreesRight;
 
     constructor() public {
+
+            bytes32 firstZero
+         = 0x290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563;
+        // i = 0
+        zeros[0] = firstZero;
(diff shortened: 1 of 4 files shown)


vaibhavchellani (approved):
Lgtm

contracts/AccountTree.sol
```diff
@@ -25,6 +25,13 @@ contract AccountTree {
     bytes32[DEPTH - BATCH_DEPTH] public filledSubtreesRight;
 
     constructor() public {
+
```
vaibhavchellani, line 28: Can we lint on this?

vaibhavchellani (approved):
LGTM!

contracts/AccountTree.sol
ChihChengLiang: sure thing 👍

PR #453, opened 2021-01-28
Fix BLS: gas computation and minor issues

ChihChengLiang:
- fix gasCost shadowing
- fix mutability
- fix input size

### DIFF
--- a/contracts/libs/BLS.sol
+++ b/contracts/libs/BLS.sol
@@ -65,7 +65,7 @@ library BLS {
         uint256 precompileGasCost = BNPairingPrecompileCostEstimator(
             COST_ESTIMATOR_ADDRESS
         )
-            .gasCost(2);
+            .getGasCost(2);
         bool callSuccess;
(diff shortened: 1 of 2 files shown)


contracts/libs/BLS.sol
```diff
@@ -114,7 +114,7 @@ library BLS {
         uint256[1] memory out;
 
         // prettier-ignore
-        uint256 precompileGasCost = BNPairingPrecompileCostEstimator(COST_ESTIMATOR_ADDRESS).gasCost(inputSize);
+        uint256 precompileGasCost = BNPairingPrecompileCostEstimator(COST_ESTIMATOR_ADDRESS).getGasCost(size + 1);
```
ChihChengLiang, line 117: This slip through from the previous PR, resulting 6x higher gas cost

where 
size = pubkeys.length
inputSize = (size + 1) * 6

kilic (approved):

PR #463, opened 2021-02-01
different style of avoiding damaging merklise input data

ChihChengLiang:
- in batch level 4 this approach costs a few more hundred gas
- in batch level 10
  - before 470927
  - after    513051

### DIFF
--- a/contracts/AccountTree.sol
+++ b/contracts/AccountTree.sol
@@ -78,21 +78,18 @@ contract AccountTree {
             "AccountTree: right set is full "
         );
 
-        bytes32[BATCH_SIZE / 2] memory buf;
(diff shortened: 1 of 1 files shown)


contracts/AccountTree.sol
```diff
@@ -78,21 +78,18 @@ contract AccountTree {
             "AccountTree: right set is full "
         );
 
-        bytes32[BATCH_SIZE / 2] memory buf;
```
ChihChengLiang, line 81: unlike the merklise libaray case, which uses a dynamic array. Here the buf, pNodes, and leafs must match each other's array size.

ChihChengLiang:
abort for inefficient memory usage

Issue #465, opened 2021-02-05
Add a deposit function in the burn auction


ChihChengLiang:
### What's wrong

It would be nice if the operator deposits a large amount of ether to the burn auction contract so that it no longer needs to send ethers every time they want to bid.

The current bid method kind of already allow this but with some limitation. If we want to deposit with the `bid` method, it requires the bid amount to beat the previous bidder.

### How can we fix it

Add a deposit method

ChihChengLiang:
Fixed in #466

PR #466, opened 2021-02-05
deposit in burn auction

ChihChengLiang:
### What's wrong

#465

TODO:

- [x] add tests
- [x] Think about the side effect combined with bid (Shuoldn't be an issue)
- [x] Think about separate the function of bid and deposit. i.e. no deposit in bid (no need for now)


vaibhavchellani (approved):
LGTM!

Issue #464, opened 2021-02-03
Reduce register pubkey batch sizes


vaibhavchellani:
The current implementation allows batch register of 16pubkeys which means that the coordinator needs to wait for 16 new create2txs before he can register the batch on-chain and process the transaction. 

This means that once the user uses create2transfer to send someone payment his normal transfers won't get processed till the create2transfer goes through.

vaibhavchellani:
Can we have multiple register batches for different sizes?

ChihChengLiang:
Yeah, that sounds like a good idea. I propose we

1. Restore the account tree batch register size back to 1024 pubkeys. The batch register was designed to onboard a lot of users.
2. Create a medium size solution for our usual use cases.

The medium size solution is a contract wrapper that call AccountTree.registerSingle serval times. So we can save the 21000 gas for base ETH1 tx.

ChihChengLiang:
Keep the status quo before the devnet

PR #467, opened 2021-02-05
Mid size insert account

ChihChengLiang:
### What's wrong

#464


ChihChengLiang:
We'll have more idea what utilities we should add after we have the devnet runnnig

PR #469, opened 2021-02-09
Add account root to log

ChihChengLiang:
### What's wrong

- We currently don't emit account tree root when submitting batch. The coordinator has no way to know what account tree root  is used to create the batch
- The msg sender in the new batch log is kind of redundant. The same information can be acquired from the view function of the batch meta.

### How are we fixing it

We remove the msg sender field and add account tree root field. So the tps for transfer/mm/c2t remains to be the same.


vaibhavchellani (approved):
lgtm

Issue #390, opened 2020-12-13
Create deploy scripts for Registering Tokens and Fee Receivers


ChihChengLiang:
### What's wrong

When we deploy the system, we want to 
- Register a list of token of interest
- Deposit for coordinators so they can collect fees for those tokens

### How can we fix it

- Create a config that has a list of addresses and the name of the token (maybe also tokenID)
- add a script that does a mass deposit for coordinators

### Problem

The deposit manager currently requires a non-zero amount. Do coordinators really need to have all the shitcoins to do the deposit?

vaibhavchellani:
> Do coordinators really need to have all the shitcoins to do the deposit?

Good question. So far there are 2 ways to create a new state leaf, deposits or create2transfer and amongst them, only one way to enter L2 with a new token i.e `deposit` so it does seem like the coordinator has to have a state leaf for every shitcoin he wants to accept fees in

ChihChengLiang:
What do you think we allow 0 amount but charge a Eth fee to deposit?

vaibhavchellani:
Ah! That's interesting, but the deposit manager won't allow that, would it?

ChihChengLiang:
Close this in favor of the new one  #476

PR #471, opened 2021-02-12
remove amount greater than 0 requirement

ChihChengLiang:
### What's wrong

As mentioned in #390, the deposit manager currently requires a non-zero amount. If people register a new token in the token registry, the coordinators then need to have an L2 state to serve that token. But the deposit manager's "amount > 0" restriction means the coordinator needs to hold some of that token to deposit.

The reason to have that restriction in the first place is to prevent the deposit function from being DoSed. But if the token is already a shitcoin, the restriction means nothing.

We could argue that paying gas is already mitigating DoS to some degree. And we can add more other measures in the future.

### DIFF
--- a/contracts/DepositManager.sol
+++ b/contracts/DepositManager.sol
@@ -118,8 +118,6 @@ contract DepositManager is DepositCore, IDepositManager {
         uint256 amount,
         uint256 tokenID
     ) external {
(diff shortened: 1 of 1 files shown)


vaibhavchellani (approved):
LGTM!

PR #473, opened 2021-02-13
Separate Float and FixedNumber

ChihChengLiang:
### What's wrong

A source of confusion for handling the value of amount and fees is we mixed the float compression and fixed number parsing/formatting in the same class DecimalCodec.

Let's expand on the fixed number and the floating number.

"[FixedNumber](https://docs.ethers.io/v5/api/utils/fixednumber/) is a fixed-width (in bits) number with an internal base-10 divisor, which allows it to represent a decimal fractional component."  

Fixed number  is how we parse and interpret the ERC20 token value. An ERC20 token TKN with decimal 18 means "1.23 TKN" is stored as the integer 1.23*10^18 on-chain. It is "ufixed128x18" in the notion of [fixedFormat](https://docs.ethers.io/v5/api/utils/fixednumber/#FixedFormat--strings) as it is unsigned, taking 128 bits, and having 18 decimals.

Floating Number is the way we compress the transfer amount and fee values to save gas in calldata. We compress the fixed number value from 128 bits to the 16 bits floating number. To get the idea of the gas costs saving, it [costs](https://github.com/ethereum/go-ethereum/blob/master/params/protocol_params.go) 4 gas per zero bytes and 16 gas per non-zero bytes in the calldata.

We used to avoid the term compress/decompress because it indicates loss (in precision). In terms of the onchain operation, the compression/decompression is lossless. But it's lossy in terms of the value you want to send. For example, you can't send "1.2345" TKN, you must round it down to  "1.234" TKN. See `test/fast/decimal.test.ts` for more examples.
I would argue if the whole purpose of the floating number is to reduce the size of the data, then calling it compression should be fine.

### How are we fixing it

We separate the Fixed number operations and the floating-point compressions by replacing the DecimalCodec class with the ERC20 class and the float class.

Note: No contract change.

### DIFF
--- a/test/fast/decimal.test.ts
+++ b/test/fast/decimal.test.ts
@@ -1,26 +1,40 @@
 import { assert } from "chai";
-import { USDT } from "../../ts/decimal";
+import { float16, USDT } from "../../ts/decimal";
 
 describe("Decimal", () => {
-    const goodCases: number[] = [0, 1, 10000, 12.13, 0.1234, 18690000000];
-
     it("Get's some property right", () => {
-        assert.equal(USDT.bytesLength, 2);
+        assert.equal(float16.bytesLength, 2);
     });
 
-    it("casts values", () => {
-        for (const value of goodCases) {
+    it("Rounding values", () => {
+        const losslessCases: number[] = [
+            0,
+            1,
+            10000,
+            12.13,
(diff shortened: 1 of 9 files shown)


vaibhavchellani (approved):
LGTM!

PR #474, opened 2021-02-13
Add deposit subtreeID

### LINKED ISSUE #472: Deposit subtree needs an unique identifier
### What's wrong

The deposit subtree has no unique identifier. This has problems:

- The client has no way to know which deposit root emitted from DepositSubTreeReady has been processed or reorged.
- The client has no way to know if a subtree root has been finalized in the Rollup.sol.

Note that we can not use the subtree root as a unique identifier, because it's highly probable to deposit the same state leaves with the same pubkeyIDs, tokens, and balances.

### How can we fix it

- The variable `front` in the SubtreeQueue contract looks like a suitable choice for the unique identifier. It's ever-increasing and not duplicating. Return the number `front` in `dequeue` function.
- Add the unique identifier in event IDepositManager::DepositSubTreeReady.
- Add the unique identifier in event Rollup::DepositsFinalised.

ChihChengLiang:
Fix #472.

Note that #472 missed the enqueue part.

### DIFF
--- a/contracts/DepositManager.sol
+++ b/contracts/DepositManager.sol
@@ -23,11 +23,13 @@ contract SubtreeQueue {
     uint256 public front = 1;
     uint256 public back = 0;
 
-    function enqueue(bytes32 subtreeRoot) internal returns (uint256 subtreeID) {
+    event DepositSubTreeReady(uint256 subtreeID, bytes32 subtreeRoot);
+
+    function enqueue(bytes32 subtreeRoot) internal {
         uint256 _back = back + 1;
         back = _back;
         queue[_back] = subtreeRoot;
-        return _back;
+        emit DepositSubTreeReady(_back, subtreeRoot);
     }
 
     function dequeue()
@@ -54,10 +56,7 @@ contract DepositCore is SubtreeQueue {
 
     uint256 public paramMaxSubtreeSize = 2;
 
-    function insertAndMerge(bytes32 depositLeaf)
-        internal
-        returns (uint256 subtreeID, bytes32 readySubtree)
-    {
+    function insertAndMerge(bytes32 depositLeaf) internal {
         depositCount++;
         uint256 i = depositCount;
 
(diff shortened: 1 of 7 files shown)


contracts/DepositManager.sol
```diff
@@ -21,16 +23,24 @@ contract SubtreeQueue {
     uint256 public front = 1;
     uint256 public back = 0;
 
+    event DepositSubTreeReady(uint256 subtreeID, bytes32 subtreeRoot);
+
     function enqueue(bytes32 subtreeRoot) internal {
-        back += 1;
-        queue[back] = subtreeRoot;
```
ChihChengLiang, line 26: the old way has a redundant SLOAD.

contracts/DepositManager.sol
```diff
@@ -21,16 +23,24 @@ contract SubtreeQueue {
...
+        uint256 subtreeID = back + 1;
+        back = subtreeID;
+        queue[subtreeID] = subtreeRoot;
+        emit DepositSubTreeReady(subtreeID, subtreeRoot);
```
ChihChengLiang, line 32: We forgot to add DepositSubTreeReady event in the `reenqueue` function. But if we add the event in the `enqueue` function, it comes with some benefits:
- we don't worry about forgetting it anywhere. 
- We don't pass the `subtreeID` and `readySubtree` all the way to the `depositFor`.

contracts/test/TestRollup.sol
```diff
@@ -7,14 +7,16 @@ import { Types } from "../libs/Types.sol";
 import { IDepositManager } from "../DepositManager.sol";
 
 contract MockDepositManager is IDepositManager {
-    event EnqueSubtree(bytes32 subtreeRoot);
```
ChihChengLiang, line 10: I should've noticed this is the sign we forget to add `DepositSubTreeReady`

vaibhavchellani (approved):
LGTM! Thanks for the quick PR!

Issue #484, opened 2021-02-22
Remove not-indexed calldata-dependent fields from events for better gas optimization


ChihChengLiang:
### What's wrong

> We recommend remove not-indexed calldata-dependent fields from events for better gas optimization.

For example, the public key here doesn't need to be logged. https://github.com/thehubbleproject/hubble-contracts/blob/4ff3955f3faea7e61593d05a24fb8fd70ca7095f/contracts/BLSAccountRegistry.sol#L13

### How can we fix it

Review and redesign events for single and batch registration.

ChihChengLiang:
closed by #489

PR #489, opened 2021-02-23
improve key registry event gas efficiency

ChihChengLiang:
### What's wrong

Address #484


ChihChengLiang:
After the update, the gas cost of 1024 key registration is reduced from 6m to 3m. The cost of logging pubkey is insane.
OTOH, public key registration happens to be a public function. That means people can wrap registration with an arbitrary contract. So if parsing calldata will fail if people do that.

Issue #455, opened 2021-01-28
Exploring `immutable` keyword to save gas


ChihChengLiang:
### What's wrong

Since we already upgrade to Solidity 0.6, there's a nice `immutable` keyword we can consider using.

Here's how it works. We usually have variables whose values won't be changed once the contract is deployed. However, their values must be determined at the deployment stage. We can't hardcode those values in the codebase, so we can't use the `constant` keyword for them.

In our use case, deployment parameters and some contract addresses that a contract depends on are such cases.

The immutable keyword replaces the variable references in the runtime bytecode with the value defined in the constructor. No need to SLOAD the variable when querying its value.

Example:
before
```
contract Rollup {
    uint256 public paramStakeAmount;

    constructor(uint256 stakeAmount){
        paramStakeAmount = stakeAmount;
    }
    function rollback() internal {
        ...
        // SLOAD paramStakeAmount for value
        uint256 slashedAmount = nActual.mul(paramStakeAmount);
        ...
    }
}
```
after
```
contract Rollup {
    uint256 public immutable paramStakeAmount;

    // deployed with stakeAmount = 10
    constructor(uint256 stakeAmount){
        paramStakeAmount = stakeAmount;
    }
    function rollback() internal {
        ...
        // paramStakeAmount is replaced with 10 in the runtime bytecode. No SLOAD
        uint256 slashedAmount = nActual.mul(paramStakeAmount);
        ...
    }
}
```

### Benefits

This saves us one SLOAD for each potential immutable referenced in a function.

### Concerns

Add a complication to reason the contract bytecode. We need to think about the side effects.

ChihChengLiang:
https://github.com/thehubbleproject/hubble-contracts/blob/4ff3955f3faea7e61593d05a24fb8fd70ca7095f/contracts/rollup/Rollup.sol#L414
> No immutable variables are used in the Rollup contract. We propose using immutables to reduce a number of SLOADs.

ChihChengLiang:
Closed by #491 and #497

PR #491, opened 2021-02-24
Add immutable to Rollup.sol/BatchManager.sol to reduce SLOAD

ChihChengLiang:
### What's wrong 

#455

### DIFF
--- a/contracts/rollup/BatchManager.sol
+++ b/contracts/rollup/BatchManager.sol
@@ -4,18 +4,21 @@ pragma experimental ABIEncoderV2;
 
 import { SafeMath } from "@openzeppelin/contracts/math/SafeMath.sol";
 import { Types } from "../libs/Types.sol";
-import { Parameters } from "./Parameters.sol";
 import { Bitmap } from "../libs/Bitmap.sol";
 import { IDepositManager } from "../DepositManager.sol";
 import { Chooser } from "../proposers/Chooser.sol";
 
-contract BatchManager is Parameters {
+contract BatchManager {
     using SafeMath for uint256;
     using Types for Types.Batch;
 
+    uint256 public immutable paramStakeAmount;
+    uint256 public immutable paramBlocksToFinalise;
+    uint256 public immutable paramMinGasLeft;
+
     // External contracts
-    IDepositManager public depositManager;
-    Chooser public chooser;
+    IDepositManager public immutable depositManager;
+    Chooser public immutable chooser;
 
     // batchID -> Batch
     mapping(uint256 => Types.Batch) public batches;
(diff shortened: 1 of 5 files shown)


contracts/rollup/BatchManager.sol
```diff
@@ -59,6 +62,20 @@ contract BatchManager is Parameters {
         _;
     }
 
+    constructor(
```
ChihChengLiang, line 65: Solidity requires the value of immutable to be assigned immediately. We can't assign the value in the inherited contracts.

contracts/rollup/Rollup.sol
```diff
@@ -13,29 +13,82 @@ import { Create2Transfer } from "../Create2Transfer.sol";
 import { BatchManager } from "./BatchManager.sol";
 import { IDepositManager } from "../DepositManager.sol";
 
-contract RollupCore is BatchManager {
+contract Rollup is BatchManager {
```
ChihChengLiang, line 16: We no longer want to separate the Rollup and RollupCore contracts now. We didn't write tests that need to override values from RollupCore.

contracts/rollup/Parameters.sol
```diff
@@ -1,9 +0,0 @@
-// SPDX-License-Identifier: MIT
-pragma solidity ^0.6.12;
-
-contract Parameters {
-    uint256 public paramStakeAmount = 0.1 ether;
-    uint256 public paramBlocksToFinalise = 0;
-    uint256 public paramMinGasLeft = 10000;
-    uint256 public paramMaxTxsPerCommit = 32;
-}
```
ChihChengLiang, line 9: I move paramX to the place they are consumed.

Benefits:
- 1 less contract.
- They were visible where they are consumed.

Downsides:
- no centralized place to show them together. 

Note:
It's fine if we consume an immutable defined in the parent contract. No SLOAD is compiled in the child contract. But Solidity still required the function's visibility to be defined as "view", not "pure".


```
// SPDX-License-Identifier: MIT
pragma solidity ^0.6.12;

contract Parent {
    uint256 immutable a;
    constructor(uint256 _a) public{
        a = _a;
    }
    function viewParent() external view returns(uint256){
        return a;  // No SLOAD
    }
}
contract Child is Parent{
    constructor(uint256 _a) Parent(_a) public{}
    
    function viewChild() external view returns(uint256){
        return a; // No SLOAD
    }
}

```

ChihChengLiang:
In terms of gas-saving, the most significant reduction is gas per batch rollback, which is from 4958 to 4354 (Rollback 249 batches is from 1234592 to 1084244 ).

That's because the paramMinGasLeft was a SLOAD but now a constant
https://github.com/thehubbleproject/hubble-contracts/blob/ddf6cb94fbb92dab9c1a3f253fcab26d6213c10b/contracts/rollup/BatchManager.sol#L85-L86

PR #497, opened 2021-02-25
Add the rest of the immutables

ChihChengLiang:
This should complete the  #455


contracts/proposers/BurnAuction.sol
```diff
@@ -50,7 +50,7 @@ contract BurnAuction is Chooser {
         public
     {
         require(
-            donationNumerator <= DONATION_DENOMINATOR,
+            _donationNumerator <= DONATION_DENOMINATOR,
```
ChihChengLiang, line 53: Scary, how did I miss this. 🙈😱

PR #507, opened 2021-03-01
implement batch handler and calldata factories

ChihChengLiang:
### What's wrong

To sync L2 txs in Rollup.sol, we need to parse the calldata of submitX functions.

### How we are fixing it

- listen to a new Batch event
- A factory creates a batch with respect to their batch type.

### DIFF
--- a/test/slow/rollup.create2Transfer.test.ts
+++ b/test/slow/rollup.create2Transfer.test.ts
@@ -15,6 +15,7 @@ import { USDT } from "../../ts/decimal";
 import { hexToUint8Array } from "../../ts/utils";
 import { Group, txCreate2TransferFactory } from "../../ts/factory";
 import { deployKeyless } from "../../ts/deployment/deploy";
+import { handleNewBatch } from "../../ts/client/batchHandler";
 
 const DOMAIN = hexToUint8Array(
     "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"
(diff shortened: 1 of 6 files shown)


test/slow/rollup.create2Transfer.test.ts
```diff
@@ -113,20 +114,24 @@ describe("Rollup Create2Transfer", async function() {
             "submitBatch execution cost",
             (await _txSubmit.wait()).gasUsed.toNumber()
         );
+        const [event] = await rollup.queryFilter(
+            rollup.filters.NewBatch(null, null, null),
+            _txSubmit.blockHash
+        );
+        const parsedBatch = await handleNewBatch(event, rollup);
```
ChihChengLiang, line 121: It's nice we can reuse the contract tests for testing client components.

ts/client/batchHandler.ts
```diff
@@ -0,0 +1,23 @@
...
+export async function handleNewBatch(
+    event: Event,
+    rollup: Rollup
+): Promise<Batch> {
+    const ethTx = await event.getTransaction();
+    const data = ethTx?.data as string;
+    const txDescription = rollup.interface.parseTransaction({ data });
+    const batchID = event.args?.batchID;
+    const batchType = event.args?.batchType;
+    const accountRoot = event.args?.accountRoot;
```
ChihChengLiang, line 14: Things could go wrong here. Think about the error handling strategy

ts/commitments.ts
```diff
@@ -264,11 +265,51 @@ export class Batch {
     }
 }
 
+export function batchFactory(
+    batchType: Usage,
+    txDescription: TransactionDescription,
+    accountRoot: string
+): Batch {
+    if (batchType == Usage.Transfer) {
```
ChihChengLiang, line 273: We are relying on the batchType here to determine which batch class to use. 

Technically txDescription gives us the function name (like 'submitTransfer') and the function signature (such as  'submitTransfer(bytes32[],uint256[2][],uint256[],bytes[])'). We can 
1. hardcode the function names as constant. But need to remember to update when the contract changes.
2. wrangling with typechain interfaces to get the correct function name even if the contract changes. But typechain is hard to wrangle.

The batchType looks more straightforward here.

ts/commitments.ts
```diff
@@ -264,11 +265,51 @@ export class Batch {
...
+    static fromCalldata(
+        txDescription: TransactionDescription,
+        accountRoot: string
+    ) {
+        const {
+            stateRoots,
+            signatures,
+            feeReceivers,
+            txss
+        } = txDescription.args;
```
ChihChengLiang, line 298: Think about error handling strategies here.

PR #513, opened 2021-03-05
Add Undoable tree

ChihChengLiang:
### What's wrong

In the state transition process, we want to get intermediate state root and witness. And we want to revert if a state transition failed in a middle step.

### How can we fix it

Create a tree class that allow us to get intermediate value and revert when desired


ChihChengLiang:
Undo this

Issue #533, opened 2021-03-16
Geth can't estimate gas for the keyless deployment


ChihChengLiang:
### What's wrong

If we run the deploy script on Geth (1.9.25 and 1.10.0), this line fails.

https://github.com/thehubbleproject/hubble-contracts/blob/ff3e64cacfdaf2c22cdc3a46ad91b3eff8fd3626/ts/deployment/keylessDeployment.ts#L54-L56

With this error message

```
[...]

PR #534, opened 2021-03-17
Remove keyless estimation in deploy time

kilic:
#533


ChihChengLiang:
I think I'll go with the #536 approach.
Thank you for the PR.

PR #536, opened 2021-03-18
revamp keyless deployer

ChihChengLiang:
### What's wrong

The only parameter that calculateKeylessDeployment really needs is "bytecode". Other parameters like providers, gas price, gas limit, and verbose can be passed on-demand or assigned with a default value.

We run calculateKeylessDeployment multiple times to get some fields in the output. These fields include contractAddress,
 keylessAccount, estimatedGasCost, alreadyDeployed, and rawDeplotmentTx. The contractAddress, keylessAccount, and rawDeplotmentTx can be derived when we have the information of bytecode, gas price, and gas limit. But to get estimatedGasCost and alreadyDeployed we need to connect a provider.

### How can we fix it

- We create a KeylessDeployer class that holds bytecode, contractAddress, keylessAccount, and rawDeplotmentTx.  
- Connected with a provider, KeylessDeployer can 
  - estimatedGasCost when we really need the number 
  - or tell us if the contract is already deployed

### DIFF
--- a/ts/deployment/deployDeployer.ts
+++ b/ts/deployment/deployDeployer.ts
@@ -14,7 +14,12 @@ export async function deployDeployer(
 ): Promise<boolean> {
     assert(signer.provider);
     const provider = signer.provider;
-    const deployer = new KeylessDeployer(deployerBytecode()).connect(provider);
+    // Need exact GAS_PRICE and GAS_LIMIT to derive correct DEPLOYER_ADDRESS
+    const deployer = new KeylessDeployer(
+        deployerBytecode(),
+        KEYLESS_DEPLOYMENT.GAS_PRICE,
+        KEYLESS_DEPLOYMENT.GAS_LIMIT
+    ).connect(provider);
 
     if (await deployer.alreadyDeployed()) {
         logAddress(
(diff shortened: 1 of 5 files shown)


ts/deployment/deployDeployer.ts
```diff
@@ -1,80 +1,42 @@
...
-    assert(KEYLESS_DEPLOYMENT.GAS_LIMIT.gte(result.estimatedGasCost));
-    assert(DEPLOYER_ADDRESS == result.contractAddress);
-    const _result = await keylessDeploy(
-        signer,
-        deployerBytecode(),
-        KEYLESS_DEPLOYMENT.GAS_PRICE,
-        KEYLESS_DEPLOYMENT.GAS_LIMIT,
-        verbose
-    );
+    assert(DEPLOYER_ADDRESS == deployer.contractAddress);
```
ChihChengLiang, line 32: We here check only DEPLOYER_ADDRESS because that's what matters.
We can't do `assert(KEYLESS_DEPLOYMENT.GAS_LIMIT.gte(await deployer.estimateGas()));` here because at the stage, we haven't funded the keylessAccount yet, so the estimateGas will fail for the insufficient fund. Hardhat seems to be fine with this, but Geth is strict 😢

PR #552, opened 2021-04-19
WIP Add solidity-coverage

jacque006:
Add coverage for contracts via `solidity-coverage` to help identify areas that need additional tests both now and in the future.

### TODO
- [ ] Coverage is not reporting statements and lines, may be due to https://github.com/sc-forks/solidity-coverage/issues/574 although I only found one test case which calls `hardhat_reset`
- [ ] Determine if tests involving gas prices will be affected. https://hardhat.org/plugins/hardhat-gas-reporter.html might be a good followup to add so we can keep an eye on it.
- [ ] Remove `skip`s/`only`s


test/fast/accountTree.test.ts
```diff
@@ -8,7 +8,7 @@ import { ZERO_BYTES32 } from "../../ts/constants";
 
 let DEPTH: number;
 let BATCH_DEPTH: number;
-describe("Account Tree", async () => {
+describe.only("Account Tree", async () => {
```
jacque006, line 11: @ChihChengLiang This ideally would fail in CI from linting, could use https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/no-exclusive-tests.md although I don't think we have eslint setup. Can we add eslint rules to run through prettier?

ChihChengLiang: Adding the no-exclusive eslint rules sounds like a good idea! Forgetting to remove "only" is one of the biggest nightmares for me. Feel free to create an issue for this.

jacque006: https://github.com/thehubbleproject/hubble-contracts/issues/554

cgewecke:
Sorry for the delay debugging this. You're seeing zero-coverage because:
+ solidity-coverage modifies your contracts (injecting trace instrumentation) and recompiles them, saving the result to artifacts
+ you consume your bytecode from typechain factories in `types/`

 You need to run typechain after solidity-coverage executes it's custom compilation. You can do this using a workflow hook in `.solcover.js`

With this config
```js
const shell = require('shelljs'); // This is a dep at solidity-coverage, no need to install separately

module.exports = {
    skipFiles: ["./test", "./deployment"],
    onCompileComplete: async function(config){
      shell.exec("typechain --target ethers-v5 './artifacts/contracts/**/!(*.dbg).json'");
    }
};
```

... you'll see:
Screen Shot 2021-04-25 at 3 27 16 PM

cgewecke:
Oh! Just remembered, there's (at least) one more potential coverage issue. SC doesn't handle `hardhat_reset` well, because that method completely re-instantiates the provider.

https://github.com/thehubbleproject/hubble-contracts/blob/7d28e78ae646c7a176fa4e8c543d3f20966b36f0/test/fast/frontends.test.ts#L28 

It should be possible to achieve the same thing using `evm_snapshot` and `evm_revert` without running into problems.

jacque006:
Thanks @cgewecke ! Things are looking much better:

Screenshot from 2021-04-26 10-26-46

Still running into some issues with tests involving gas consumption, but I was expecting that from https://github.com/sc-forks/solidity-coverage#usage-notes

Issue #553, opened 2021-04-20
Sync public key registration


ChihChengLiang:
### What's wrong

The syncer now only learns only new batch events on-chain, we need to sync new public key registration too. So that the client can validate the signature signed by the latest added keys.

### How can we fix it

Add method to parse public event in the initialSync function in the SyncerService. 

It would be similar to how we parse batches. 

- We loop a chunk of 100 eth1 blocks.
- In each chunk, we use the ethers.js's query filter function to get all the events in those blocks.
- for each event, we parse the data we want.

https://github.com/thehubbleproject/hubble-contracts/blob/42b6f3676dd9dda8ac601dae0137a1c4da87c1ba/ts/client/services/syncer.ts#L42-L55

There are two ways to register public keys. One is to register a single public key and the other is to register many public keys at once.

They emit SinglePubkeyRegistered and BatchPubkeyRegistered respectively.

Note that in those events you can't get the actual public key. You can get only the public key ID.
To get the actual public key we need to parse calldata like what we did in parsing the batch.

To keep the task size small, we could implement only the `SinglePubkeyRegistered` and log an error for the `BatchPubkeyRegistered`. We can implement `BatchPubkeyRegistered` later in the next issue.

ChihChengLiang:
Note that we use `storageManagerFactory` to add preconfigured public keys to the offchain pubkeyStorage in the integration test and in the real node. That was for convenience reason so we can have public keys offchain without syncing from the contract.

But that also causes the off-chain data inconsistent with the on-chain public key contract (blsAccountRegistry).

Once public key sync is implemented, we need to 

1. Remove the adding public key part in `storageManagerFactory`. 
2. Add the preconfigured public keys to the blsAccountRegistry in the deployment script.

PR #556, opened 2021-04-21
Sync public key registration

jacque006:
Resolves https://github.com/thehubbleproject/hubble-contracts/issues/553

~TODO~
 - [x] Finish sync integration test data & verification.
 - [x] Allow pre-set public keys to be generated and/or specified in deployment script.
 - [x] Remove Group/pubkey generation points.
 - [x] Add handler for `BatchPubkeyRegistered` and (optional) implement or add stub.

### DIFF
--- a/scripts/deploy.ts
+++ b/scripts/deploy.ts
@@ -4,21 +4,50 @@ import { DeploymentParameters } from "../ts/interfaces";
 import fs from "fs";
 import { PRODUCTION_PARAMS } from "../ts/constants";
 import { StateTree } from "../ts/stateTree";
+import { Group } from "../ts/factory";
 
-const argv = require("minimist")(process.argv.slice(2), {
-    string: ["url", "root", "key", "input", "output"]
+const {
+    url,
+    root,
+    key,
+    input,
+    output,
+    numPubkeys,
+    pubkeyMnemonic
+} = require("minimist")(process.argv.slice(2), {
+    string: [
+        "url",
+        "root",
+        "key",
+        "input",
+        "output",
+        "numPubkeys",
+        "pubkeyMnemonic"
+    ]
 });
 /*
     Note separate pubkeys with commas
     > npm run deploy -- --url http://localhost:8545 \
     --root 0x309976060df37ed6961ebd53027fe0c45d3cbbbdfc30a5039e86b2a7aa7fed6e
 
-    You can also specify a private key
-    > npm run deploy -- --key 0xYourPrivateKey 
+    You can specify an Eth1 private key
+    > npm run deploy -- --key 0xYourPrivateKey
+
+    You can specify an initial number of BLS public keys to register and their mnemonic seed (hardhat mnemonic default)
+    > npm run deploy -- --numPubkeys 32 \
+    --pubkeyMnemonic 'focus hood pipe manual below record fly pole corn remember define slice kitchen capable search'
 
     You can use a custom parameters.json
     > npm run deploy -- --input parameters.json --output ../hubbe-commander/genesis.json
 */
+
+function validateArgv() {
+    if (pubkeyMnemonic && !numPubkeys) {
+        throw new Error(
+            "numPubkeys must be specified if a pubkeyMnemonic is provided"
+        );
+    }
+}
 
 function getDefaultGenesisRoot(parameters: DeploymentParameters) {
     const stateTree = StateTree.new(parameters.MAX_DEPTH);
@@ -28,22 +57,49 @@ function getDefaultGenesisRoot(parameters: DeploymentParameters) {
 }
 
 async function main() {
+    validateArgv();
+
     const provider = new ethers.providers.JsonRpcProvider(
-        argv.url ?? "http://localhost:8545"
+        url ?? "http://localhost:8545"
     );
-    const signer = argv.key
-        ? new ethers.Wallet(argv.key).connect(provider)
+    const signer = key
+        ? new ethers.Wallet(key).connect(provider)
         : provider.getSigner();
 
-    const parameters = argv.input
-        ? JSON.parse(fs.readFileSync(argv.input).toString())
+    const parameters = input
+        ? JSON.parse(fs.readFileSync(input).toString())
         : PRODUCTION_PARAMS;
 
-    parameters.GENESIS_STATE_ROOT =
-        argv.root || getDefaultGenesisRoot(parameters);
+    parameters.GENESIS_STATE_ROOT = root || getDefaultGenesisRoot(parameters);
     console.log("Deploy with parameters", parameters);
 
-    await deployAndWriteGenesis(signer, parameters, argv.output);
+    const { blsAccountRegistry } = await deployAndWriteGenesis(
+        signer,
+        parameters,
+        output
+    );
+
+    if (numPubkeys) {
+        console.log(
+            `Registering ${numPubkeys} pubkeys. Custom mnemonic: ${!!pubkeyMnemonic}`
+        );
+        const group = Group.new({
+            n: pubkeyMnemonic,
+            mnemonic: pubkeyMnemonic
+        });
+        // Convert this to batch register once implemented
+        for (const user of group.userIterator()) {
+            await blsAccountRegistry.register(user.pubkey);
+        }
+    }
 }
 
-main();
+main()
+    .then(() => {
+        console.log("Deployment complete");
+        process.exit(0);
+    })
+    .catch(err => {
+        console.error(err);
+        process.exit(1);
+    });
--- a/test/client/integration.test.ts
+++ b/test/client/integration.test.ts
@@ -9,11 +9,14 @@ import {
 } from "../../ts/client/features/transfer";
 import { SyncerService } from "../../ts/client/services/syncer";
 import { PRODUCTION_PARAMS } from "../../ts/constants";
+import { USDT } from "../../ts/decimal";
 import { deployAll } from "../../ts/deploy";
 import { deployKeyless } from "../../ts/deployment/deploy";
 import { Group, storageManagerFactory } from "../../ts/factory";
 import { Genesis } from "../../ts/genesis";
 import * as mcl from "../../ts/mcl";
+import { Pubkey } from "../../ts/pubkey";
+import { State } from "../../ts/state";
 
 describe("Client Integration", function() {
     it("run", async function() {
@@ -22,15 +25,45 @@ describe("Client Integration", function() {
         const provider = signer.provider as providers.Provider;
         const genesisEth1Block = await provider.getBlockNumber();
         await deployKeyless(signer, false);
+        const storagePacker = await storageManagerFactory();
+        const storageSyncer = await storageManagerFactory();
+
+        // Setup pubkeys, state for packer & syncer
+        const tokenID = 1;
+        const initialBalance = USDT.fromHumanValue("100.12");
         const group = Group.new({ n: 32 });
-        const storagePacker = await storageManagerFactory(group);
-        const storageSyncer = await storageManagerFactory(group);
+        for (const user of group.userIterator()) {
+            const state = State.new(
+                user.pubkeyID,
+                tokenID,
+                initialBalance.l2Value,
+                0
+            );
+            // Setup packer L2 storage
+            await storagePacker.pubkey.update(
+                user.pubkeyID,
+                new Pubkey(user.pubkey)
+            );
+            await storagePacker.state.update(user.stateID, state);
+
+            // Setup syncer L2 state
+            // Replace with L1 deposits once implemented
+            await storageSyncer.state.update(user.stateID, state);
+        }
+        await storagePacker.pubkey.commit();
+        await storagePacker.state.commit();
+        await storageSyncer.state.commit();
 
         const parameters = PRODUCTION_PARAMS;
         parameters.USE_BURN_AUCTION = false;
         parameters.GENESIS_STATE_ROOT = storagePacker.state.root;
 
         const contracts = await deployAll(signer, parameters);
+        for (const user of group.userIterator()) {
+            // Setup L1 pubkeys
+            await contracts.blsAccountRegistry.register(user.pubkey);
+        }
+
         const appID = await contracts.rollup.appID();
         group.setupSigners(arrayify(appID));
 
@@ -58,6 +91,8 @@ describe("Client Integration", function() {
 
         const syncService = new SyncerService(apiSyncer);
         await syncService.initialSync();
+
         assert.equal(storageSyncer.state.root, storagePacker.state.root);
+        assert.equal(storageSyncer.pubkey.root, storagePacker.pubkey.root);
     }).timeout(300000);
 });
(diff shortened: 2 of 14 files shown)


ts/client/services/syncer.ts
```diff
@@ -81,13 +120,46 @@ export class SyncerService {
...
 
+    getPubkeyFromTxnData(data: string): [string, string, string, string] {
+        // TODO Is there a better way to do this?
+        const [
+            pubkey
+        ] = this.api.contracts.blsAccountRegistry.interface.decodeFunctionData(
+            "register",
+            data
+        );
+        return pubkey;
```
jacque006, line 131: @ChihChengLiang I'm all :ear:s if you know of a better way to do this

test/client/integration.test.ts
```diff
@@ -45,6 +45,11 @@ describe("Client Integration", function() {
         const apiPacker = CoreAPI.new(storagePacker, genesis, provider, signer);
         const apiSyncer = CoreAPI.new(storageSyncer, genesis, provider, signer);
 
+        const groupToRegister = Group.new({ n: 1 });
+        await apiPacker.contracts.blsAccountRegistry.register(
+            groupToRegister.getUser(0).pubkey
+        );
```
jacque006, line 51: Is there a preferred abstraction/way of doing this test data setup, suck as the packer command below?

ts/client/services/syncer.ts
ChihChengLiang: What do you think about using
```
this.api.contracts.blsAccountRegistry.interface.parseTransaction({data})
```
- Pros: no magic string "register"
- Cons: It's implicit. Readers don't know we want to parse "register" here.

test/client/integration.test.ts
ChihChengLiang: Haven't developed an opinion on this. I now see some options here, but leave you to decide. It's also fine to do a quick and dirty one here as long as it works.

- Add a register method in Group. `await group.registerKeyOneByOne()`
  - pros: looks straight forward
  - cons: need to pass API to group somehow. Group knows too much.
- Add registerKeyOneByOne to API, which takes group as an argument. `await api.registerKeyOneByOne(group)`
  - pros: Group stays agnostic to API
  - cons: API should reflect the node's functionality. When we implement Create2Transfer the coordinator does need to help to register a public key for a user. But in the test here we are using this method to do test data setup, which is not a functionality of the node, so it might confuse people.
- Add a third class/function to do test data setup `await registerPublicKey(api, group)`
  - pros: no extra responsibilities for API or Group
  - cons: It looks like we are just extracting a function for these two lines of code, not improving readability a lot. 

Let me also expand on TransferPackingCommand too. The coordinator needs to pack Deposit, Transfer, MassMigration, or Create2Transfer batches when they win a slot to propose. But each type of batch depends on some objects to pack. To unify the packing interface, we created BatchPackingCommand that has the packAndSubmit method. So that in the packer service, packer can select one of the commands and fire the packAndSubmit method without worrying about which objects the commands depend on.

ts/client/coreAPI.ts
```diff
@@ -98,7 +98,7 @@ export class CoreAPI implements ICoreAPI {
     async getBlockNumber() {
         return await this.provider.getBlockNumber();
     }
-    async getlatestBatchID() {
+    async getLatestBatchID() {
```
ChihChengLiang, line 101: Nice catch!

ts/client/services/syncer.ts
jacque006: I think that's a much less brittle approach. I can add a comment indicating that we are getting the pubkey call data from the transaction data.

Also, is there a way we can do this in bulk? If so I can re-work some of this functionality to grab transaction data 10 at a time to reduce the number of RPC calls to eth node.

test/client/integration.test.ts
jacque006: Thanks for the packing command explanation, that makes more sense now.

I think if this data setup gets more complicated I will go with the third option but for now I will [KISS](https://en.wikipedia.org/wiki/KISS_principle)

ts/client/services/syncer.ts
ChihChengLiang: Not sure if we can query eth1 tx in bulk. We now assume the coordinator will maintain their own eth1 node instead of relying on Infura so that we don't need to worry about the RPC call quota.

jacque006: Ok, sounds good. I mainly was thinking from a network i/o perspective, doing a batched lookup will be faster/cheaper than many transaction lookups.

ts/utils.ts
```diff
@@ -147,3 +147,11 @@ export function prettyHex(hex: string) {
...
     return `${hexNo0x.slice(0, 6)}…${hexNo0x.slice(-6)}`;
 }
+
+export function chunkArray<T>(arr: T[], chunkSize: number): T[][] {
+    const chunkArr = [];
+    while (arr.length) {
+        chunkArr.push(arr.splice(0, chunkSize));
+    }
+    return chunkArr;
+}
```
jacque006, line 157: If we want a more battle hardened version of this, we can bring in `lodash`'s [_.chunk](https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L6839)

.github/workflows/nodejs.yml
```diff
@@ -28,6 +28,7 @@ jobs:
     - run: npm ci
     - run: npm run lint && npm run solhint
     - run: npm run generate
+    - run: npm run tsc
```
jacque006, line 31: It didn't look like we were running typescript transpiler in CI so added this. If this is being done somewhere else can remove this step.

scripts/deploy.ts
```diff
@@ -39,6 +39,12 @@ async function main() {
         ? JSON.parse(fs.readFileSync(argv.input).toString())
         : PRODUCTION_PARAMS;
 
+    // TODO public keys passed in here
+    // pass in mnemonic seed
+    // pass in number of users
+    // argv.numUsers;
+    // argv.mnemonicSeed;
```
jacque006, line 46: @ChihChengLiang Spaced on how exactly we want to handle deployed public keys, can sync up with you to determine implementation.

test/client/integration.test.ts
```diff
@@ -45,6 +48,21 @@ describe("Client Integration", function() {
...
+            const initialBalance = USDT.fromHumanValue("100.12");
+            const state = State.new(
+                user.pubkeyID,
+                tokenID,
+                initialBalance.l2Value,
+                0
+            );
+            await apiSyncer.l2Storage.state.update(user.stateID, state);
+        }
+        await apiSyncer.l2Storage.state.commit();
```
jacque006, line 64: @ChihChengLiang This was needed to setup syncer state for transaction below since it was removed from storage factory. Do we already the functionality in the node to setup initial state/balances or is that future work?

ts/client/features/transfer.ts
```diff
@@ -465,9 +471,39 @@ export class TransferPool implements ITransferPool {
...
+                this.initialBalance.l2Value,
+                0
+            );
+            await this.engine.update(user.stateID, state);
+            await this.pubkey.update(user.pubkeyID, new Pubkey(user.pubkey));
+        }
+        await this.engine.commit();
+        await this.pubkey.commit();
+
+        await this.setTokenID();
```
jacque006, line 503: I moved these to `SimulatorPool` as it seemed the right place for test data setup.

ts/client/features/transfer.ts
```diff
@@ -465,9 +471,39 @@ export class TransferPool implements ITransferPool {
...
+
+    private readonly pubkey: PubkeyStorageEngine;
+
+    constructor(group: Group, storageManager: StorageManager) {
+        super(group, storageManager.state);
+        this.pubkey = storageManager.pubkey;
+
+        // TODO Make these configurable?
+        this.tokenID = 1;
+        this.initialBalance = USDT.fromHumanValue("100.12");
```
jacque006, line 486: @ChihChengLiang Will these every need to be configured from these defaults for future test cases? If so I can add an options object to the pool constructor.

ts/client/features/transfer.ts
```diff
@@ -480,6 +516,7 @@ export class SimulatorPool extends OffchainTransferFactory
 
     getNextPipe() {
         const source = this.genTx();
+        // TODO Remove this check?
         if (this.tokenID === undefined) throw new Error("tokenID not set");
```
jacque006, line 520: If `tokenID` is always set in constructor (https://github.com/thehubbleproject/hubble-contracts/pull/556/files#diff-f8e595b0b3744ed7278b7e180945494868c328b812257e8f93251f538bef57d3R485) we should be able to remove this safety check

ts/client/node.ts
```diff
@@ -57,15 +57,21 @@ export class HubbleNode {
         const signer = provider.getSigner();
         const { parameters, auxiliary } = genesis;
 
-        const group = Group.new({ n: 32, domain: arrayify(auxiliary.domain) });
-        const storageManager = await storageManagerFactory(group, {
-            stateTreeDepth: parameters.MAX_DEPTH
+        // TODO Do we need to apply auxiliary.domain provided to state storage for bls signing?
+        // const group = Group.new({ n: 32, domain: arrayify(auxiliary.domain) });
```
jacque006, line 61: @ChihChengLiang

.github/workflows/nodejs.yml
jacque006: Not needed, remove

tsconfig.json
```diff
@@ -13,10 +13,11 @@
     "lib": ["es2015", "esnext.asynciterable"],
     "sourceMap": true,
     "typeRoots": ["./node_modules/@types", "./types"],
-    "types": ["node"]
+    "types": ["node"],
+    "outDir": "./build"
   },
   "include": ["**/*.ts"],
-  "exclude": ["node_modules", "example", "build"],
+  "exclude": ["node_modules", "example", "build", "test"],
```
jacque006, line 20: Remove `test` exclusion.

ts/utils.ts
jacque006: Switched to lodash.

test/client/integration.test.ts
jacque006: We do not have deposit syncing yet, used L2 setup

ts/client/features/transfer.ts
jacque006: `as it seemed the right place for test data setup`... And I was wrong. Moved back to client integration test.

jacque006:
@ChihChengLiang Ready for your review & feedback.

test/client/integration.test.ts
```diff
@@ -45,6 +48,36 @@ describe("Client Integration", function() {
...
+            );
+            // Setup packer L2 storage
+            await storagePacker.pubkey.update(
+                user.pubkeyID,
+                new Pubkey(user.pubkey)
+            );
+            await storagePacker.state.update(user.stateID, state);
+
+            // Setup syncer L1 pubkeys
+            await apiSyncer.contracts.blsAccountRegistry.register(user.pubkey);
```
ChihChengLiang, line 70: The reference to the syncer here might be confusing. 

In the test case, we assume the syncer is agnostic to the public key before sync.

```suggestion
            // Setup L1 pubkeys
            await contracts.blsAccountRegistry.register(user.pubkey);
```

scripts/deploy.ts
```diff
@@ -4,22 +4,51 @@ import { DeploymentParameters } from "../ts/interfaces";
...
-const argv = require("minimist")(process.argv.slice(2), {
-    string: ["url", "root", "key", "input", "output"]
+const {
+    url,
+    root,
+    key,
+    input,
+    output,
+    numPubkeys,
+    pubkeyMnemonic
```
ChihChengLiang, line 16: Nice cleaning up these

ts/client/services/syncer.ts
```diff
@@ -81,13 +139,47 @@ export class SyncerService {
...
+    singlePubkeyRegisteredListener = async (pubkeyID: null, event: Event) => {
+        return this.handleSinglePubkeyRegistered(event);
+    };
+
+    batchPubkeyRegisteredListener = async (
+        startID: null,
+        endID: null,
+        event: Event
+    ) => {
+        return this.handleSinglePubkeyRegistered(event);
```
ChihChengLiang, line 182: Should be handleBatchPubkeyRegistered here?
```suggestion
        return this.handleBatchPubkeyRegistered(event);
```

ts/client/services/syncer.ts
```diff
@@ -81,13 +139,47 @@ export class SyncerService {
...
+    }
+
     newBatchListener = async (
         batchID: null,
         accountRoot: null,
         batchType: null,
         event: Event
     ) => {
-        return await this.handleNewBatch(event);
+        return this.handleNewBatch(event);
```
ChihChengLiang, line 170: The function is async, why we don't await here?

test/client/integration.test.ts
```diff
@@ -22,20 +25,20 @@ describe("Client Integration", function() {
...
         await deployKeyless(signer, false);
-        const group = Group.new({ n: 32 });
-        const storagePacker = await storageManagerFactory(group);
-        const storageSyncer = await storageManagerFactory(group);
+        const storagePacker = await storageManagerFactory();
+        const storageSyncer = await storageManagerFactory();
 
         const parameters = PRODUCTION_PARAMS;
         parameters.USE_BURN_AUCTION = false;
         parameters.GENESIS_STATE_ROOT = storagePacker.state.root;
```
ChihChengLiang, line 33: The `GENESIS_STATE_ROOT` here depends on storagePacker's genesis states, so that we should setup  storagePacker before setting up `GENESIS_STATE_ROOT`.

We should move stuff in line 51~80 to before line 31

The error is not failing the test because #559

ChihChengLiang (commented):
Great work! Here's my first round of review.

ts/client/services/syncer.ts
jacque006: :facepalm: Yes, good catch

ts/client/services/syncer.ts
jacque006: When chaining `async` functions, returning an `awaited` async function is mostly equivalent to just returning the async function.

Screenshot from 2021-04-22 14-45-46

This is also true when it throws/rejects, as it will bubble up to the top await.
That being said, all these functions return void values so we can just await rather than return. Will update.

See also this silliness:
Screenshot from 2021-04-22 14-35-15

ts/client/features/transfer.ts
```diff
@@ -467,6 +467,7 @@ export class SimulatorPool extends OffchainTransferFactory
     implements ITransferPool {
     private tokenID?: number;
     private feeReceiverID?: number;
+
```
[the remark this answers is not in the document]
jacque006: My bad, missed this. Will remove and rebase.

jacque006:
@ChihChengLiang Ready for re-review

ts/client/services/syncer.ts
ChihChengLiang: Thanks for the explanation. That makes sense to me now.

ChihChengLiang (approved):
Looks good to me! That's some awesome work!

Issue #555, opened 2021-04-20
Circular Dependency Warning


jacque006:
When running tests (`npm test`) and some scripts, the following NodeJS warning appears:
```bash
(node:41227) Warning: Accessing non-existent property 'INVALID_ALT_NUMBER' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
```
Likely an issue with a dependency that needs to be updated or a bad import cycle in code.

ChihChengLiang:
Yeah, this is an issue I've been wanting to address. NodeJS seems to allow some degree of circular dependency and raises no error for it.
Do you know any tool that we can detect the circular dependency? Or what's the best practice to do the import?

jacque006:
Adding the `--trace-warnings` as a pass through arg (i.e. `npm test -- --trace-warnings`) should give the import stack trace. Also of note is that this mainly shows up for NodeJS 14 as they added better detection/warnings for circular deps. 

Found https://github.com/antlr/antlr4/issues/2834 which looks to be the culprit. Just need to find the lib we are using that has that as a dep and update it if possible.

jacque006:
Actual issue: https://github.com/prettier-solidity/prettier-plugin-solidity/issues/335

PR #560, opened 2021-04-22
Update prettier-plugin-solidity

jacque006:
Resolves https://github.com/thehubbleproject/hubble-contracts/issues/555

This update changes the formatting on a number of contracts. Not sure if this is desired. May be able to configure prettier solidity rules to keep old formatting with new lib version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "mcl-wasm": "^0.4.5",
     "minimist": "^1.2.5",
     "prettier": "^1.19.1",
-    "prettier-plugin-solidity": "^1.0.0-alpha.59",
+    "prettier-plugin-solidity": "1.0.0-beta.10",
     "solhint": "^3.3.3",
     "ts-node": "^8.8.1",
     "typechain": "^2.0.1",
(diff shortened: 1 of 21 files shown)


package.json
```diff
@@ -43,7 +43,7 @@
     "mcl-wasm": "^0.4.5",
     "minimist": "^1.2.5",
     "prettier": "^1.19.1",
-    "prettier-plugin-solidity": "^1.0.0-alpha.59",
+    "prettier-plugin-solidity": "1.0.0-beta.10",
```
jacque006, line 46: Note when dealing with `alpha/beta/rc` package versions it is generally a good idea to pin to an exact version until the release is stabilized. Similar to libs which have an unstable major version (i.e. `0.x.y`) and may have breaking changes between minor versions.

ChihChengLiang (approved):
LGTM, great work fixing the circular dependency issue. It creates lots of diffs in the contracts but I think that's fine.

package.json
ChihChengLiang: I see, will follow this practice in the future.

Issue #517, opened 2021-03-10
Event DepositQueued can be simplified


ChihChengLiang:
### What's wrong

Event DepositQueued tells the client 3 pieces of information of a freshly created state: pubkeyID, tokenID, and l2Amount. The nonce is always 0, so not required.

The current DepositQueued contains fields of unuseful and redundant information. pubkey 1 word and encodedState is probably 5 words (1 for the length of bytes and 4 for the contents of state)

https://github.com/thehubbleproject/hubble-contracts/blob/f7666140ef38b72b73d2ddd28825a8bf6705bbed/contracts/DepositManager.sol#L149

### How can we fix it

Ideally make it `event DepositQueued(uint32 pubkeyID, uint32 tokenID, uint128 l2Amount);` so it can be packed in 32 bytes.

jacque006:
Note that in [depositFor](https://github.com/thehubbleproject/hubble-contracts/blob/master/contracts/DepositManager.sol#L131-L159), which emits `DepositQueued` the new event args are all `uint256` in function.

@ChihChengLiang given that increased size, do we still want to update the `DepositQueued` event? Or am I missing something here?

ChihChengLiang:
Ah sorry, my description is confusing. 

The existing `DepositQueued` emits unnecessary information. It's ok we still use uint256 in the `DepositQueued` event. But we shouldn't have the encodedState field as what it is now. The reason is encodedState is the pubkeyID+tokenID+balance+nonce, where pubkeyID is duplicated information and nonce is unhelpful as it should always be 0.

PR #562, opened 2021-04-23
Simplify DepositQueued event

jacque006:
Resolves https://github.com/thehubbleproject/hubble-contracts/issues/517

### DIFF
--- a/test/fast/deposit.test.ts
+++ b/test/fast/deposit.test.ts
@@ -1,6 +1,6 @@
 import { ethers } from "hardhat";
 import { assert } from "chai";
-import { constants } from "ethers";
+import { constants, Event } from "ethers";
 import { solidityKeccak256 } from "ethers/lib/utils";
 import { allContracts } from "../../ts/allContractsInterfaces";
 import { TESTING_PARAMS } from "../../ts/constants";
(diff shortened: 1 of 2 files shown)


test/fast/deposit.test.ts
```diff
@@ -87,6 +87,19 @@ describe("DepositManager", async function() {
...
     it("should allow depositing 2 leaves in a subtree and merging it", async function() {
+        const getStateFromEvent = (evt: Event) => {
+            if (!evt.args) {
+                throw new Error("DepositQueued event missing args");
+            }
+            const { pubkeyID, tokenID, l2Amount } = evt.args;
+            return State.new(
+                pubkeyID.toNumber(),
+                tokenID.toNumber(),
+                l2Amount.toNumber(),
```
ChihChengLiang, line 98: l2Amount could be so large that beyonds `Number.MAX_SAFE_INTEGER`, we should keep it as bigNumber.
```suggestion
                l2Amount,
```

ChihChengLiang (approved):
LGTM after the commented issue is fixed.

test/fast/deposit.test.ts
jacque006: Good catch, and good to know that part of state needs to allow for very large numbers.

PR #569, opened 2021-04-27
Sync DepositQueued event

jacque006:
- Add `EventSyncer` interface to abstract and separate out logic that syncs L1 event to L2.
- Implement `DepositQueued` event syncing.
- Typo fixes and other minor cleanup.

~TODO~
- [x] Move `singlePubkeyRegistered`, `batchPubkeyRegistered` to `EventSyncer` pattern.
- [x] Implement event syncing for `DepositQueued`, `DepositFinalized` events.
~Complete sync integration test, implement `MemoryEngine.updateBatch`.~ https://github.com/thehubbleproject/hubble-contracts/issues/570

### DIFF
--- a/ts/client/services/syncer.ts
+++ b/ts/client/services/syncer.ts
@@ -1,13 +1,13 @@
 import { Event, EventFilter } from "@ethersproject/contracts";
 import { chunk } from "lodash";
-import { Rollup } from "../../../types/ethers-contracts/Rollup";
 import { BlsAccountRegistry } from "../../../types/ethers-contracts/BlsAccountRegistry";
-import { Usage } from "../../interfaces";
-import { BatchHandlingContext } from "../contexts";
 import { CoreAPI } from "../coreAPI";
 import { nodeEmitter, SyncCompleteEvent } from "../node";
 import { PubkeyStorageEngine } from "../storageEngine";
 import { Pubkey } from "../../pubkey";
+import { EventSyncer } from "./events/interfaces";
+import { NewBatchEventSyncer } from "./events/newBatch";
+import SequentialCompositeEventSyncer from "./events/sequentialCompositeEventSyncer";
 
 export enum SyncMode {
     INITIAL_SYNCING,
@@ -16,19 +16,15 @@ export enum SyncMode {
 
 export class SyncerService {
     private mode: SyncMode;
-    private newBatchFilter: EventFilter;
     private singlePubkeyRegisteredFilter: EventFilter;
     private batchPubkeyRegisteredFilter: EventFilter;
-    private batchHandlingContext: BatchHandlingContext;
-    private rollup: Rollup;
     private accountRegistry: BlsAccountRegistry;
     private pubkeyStorage: PubkeyStorageEngine;
+    private readonly events: EventSyncer;
 
     constructor(private readonly api: CoreAPI) {
         this.mode = SyncMode.INITIAL_SYNCING;
-        this.rollup = this.api.rollup;
         this.accountRegistry = this.api.contracts.blsAccountRegistry;
-        this.newBatchFilter = this.rollup.filters.NewBatch(null, null, null);
         this.singlePubkeyRegisteredFilter = this.accountRegistry.filters.SinglePubkeyRegistered(
             null
         );
@@ -36,8 +32,13 @@ export class SyncerService {
             null,
             null
         );
-        this.batchHandlingContext = new BatchHandlingContext(api);
         this.pubkeyStorage = this.api.l2Storage.pubkey;
+
+        this.events = new SequentialCompositeEventSyncer([
+            // Note: Ordering here is important for initial syncs.
+            // Pubkey syncs need to happen before batch syncs, etc.
+            new NewBatchEventSyncer(api)
+        ]);
     }
 
     getMode() {
@@ -48,7 +49,9 @@ export class SyncerService {
         await this.initialSync();
         nodeEmitter.emit(SyncCompleteEvent);
         this.mode = SyncMode.REGULAR_SYNCING;
-        this.rollup.on(this.newBatchFilter, this.newBatchListener);
+
+        this.events.listen();
+
         this.accountRegistry.on(
             this.singlePubkeyRegisteredFilter,
             this.singlePubkeyRegisteredListener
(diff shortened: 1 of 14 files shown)


ts/client/services/events/depositFinalised.ts
```diff
@@ -0,0 +1,63 @@
...
+        }
+        const { subtreeID, depositSubTreeRoot, pathToSubTree } = event.args;
+        // TODO Properly convert types.
+        return { subtreeID, depositSubTreeRoot, pathToSubTree };
+    }
+
+    private async handleDepositFinalised(event: Event) {
+        // const { depositSubTreeRoot } = this.getEventArgs(event);
+
+        // TODO This looks like it is already being handled in newBatch event syncer via DepositHandlingStrategy:applyBatch
```
jacque006, line 50: @ChihChengLiang Can you confirm this is correct? Since the deposit batch is already being handled in the strategy with the correct subtree pop (`deposit.ts:109`) and that event (`NewBatch`) is emitted after the `DepositFinalised` event  in `submitDeposit`.

test/client/integration.test.ts
```diff
@@ -45,24 +49,60 @@ describe("Client Integration", function() {
...
+                    tokenID,
+                    zeroNonce: true
+                });
+                const depositBatch = new BodylessCommitment(
+                    stateTree.root
+                ).toBatch();
+                return depositBatch.proofCompressed(0);
+            },
+            Promise.resolve(getGenesisProof(stateTree.root))
+        );
```
jacque006, line 105: Per the TODO, I think I found a simpler way to do this which I am looking into.

ts/client/services/events/depositFinalised.ts
ChihChengLiang: Right the comment is correct. So we don't even need DepositFinalisedEventSyncer.

The event `DepositFinalised` always comes with the `NewBatch` event of the deposit. So we don't need an extra listener for it.

Just for extra information. In `applyBatch` we apply the state transition for the batch. There are two ways we acquire a batch, 

1. A batch might be acquired during the sync. We parse the `DepositFinalised` event from it in `handleNewBatch`.
2. A batch might come from the one the packer prepared to submit. We decided to apply the local one directly instead of sync it again on chain.

test/client/integration.test.ts
```diff
@@ -45,24 +54,88 @@ describe("Client Integration", function() {
...
         }
         await storagePacker.pubkey.commit();
         await storagePacker.state.commit();
-        await storageSyncer.state.commit();
 
         const parameters = PRODUCTION_PARAMS;
         parameters.USE_BURN_AUCTION = false;
-        parameters.GENESIS_STATE_ROOT = storagePacker.state.root;
+        // TODO Check that this change is correct and makes sense.
+        parameters.GENESIS_STATE_ROOT = storageSyncer.state.root;
```
jacque006, line 64: I believe this change is correct. Until we implement genesis sync, this start state should be the syncer, not the packer.

test/client/integration.test.ts
```diff
@@ -45,24 +54,88 @@ describe("Client Integration", function() {
...
+                    stateTree.root,
+                    context
+                );
+                const depositBatch = new ConcreteBatch([commitment]);
+                const depositProof = depositBatch.proofCompressed(0);
+
+                return depositProof;
+            },
+            Promise.resolve(genesisProof)
+        );
```
jacque006, line 127: Most if not all of this can be replaced once we have the deposit packer/proposer functionality in place.

ts/client/storageEngine/memoryEngine.ts
```diff
@@ -84,6 +84,7 @@ export class MemoryEngine<Item extends Hashable>
         depth: number,
         items: Item[]
     ): Promise<void> {
+        // TODO Jake C-T implement
```
jacque006, line 87: @ChihChengLiang Any thoughts/tips on how this should be implemented? Used in the deposit batch sync `ts/client/features/deposit.ts:122`

test/client/integration.test.ts
ChihChengLiang: It depends on what we want to have in the genesis state tree.
Before we have a functioning deposit feature, we are not able to introduce a new state leaf into the state tree after the genesis. That's why we created 32 state leaves in the genesis state tree.

Since both the syncer and the packer have these 32 leaves, so it doesn't matter we use storagePacker or storageSyncer here. They should have the same state tree root.

I see you removed the setup of storageSyncer. But that would break something because now syncer and the packer has different genesis state.

ts/client/storageEngine/memoryEngine.ts
ChihChengLiang: Sorry, I forget this API is unimplemented. I've written an issue in #570

test/client/integration.test.ts
jacque006: Going to continue this work in a followup PR

test/client/integration.test.ts
jacque006: Reverted, will discuss how we want to do this offline with @ChihChengLiang

jacque006:
@ChihChengLiang Ready for your review

ts/client/services/syncer.ts
```diff
@@ -16,28 +14,19 @@ export enum SyncMode {
...
-        this.batchHandlingContext = new BatchHandlingContext(api);
-        this.pubkeyStorage = this.api.l2Storage.pubkey;
+
+        this.events = new SequentialCompositeEventSyncer([
+            // Note: Ordering here is important for initial syncs.
+            // Pubkey syncs need to happen before batch syncs, etc.
+            new SinglePubkeyRegisteredEventSyncer(api),
+            new BatchPubkeyRegisteredEventSyncer(api),
+            new DepositQueuedEventSyncer(api),
+            new NewBatchEventSyncer(api)
```
ChihChengLiang, line 28: No action to take here. I'm commenting on why in that 100 eth1 blocks interval, we can process all the deposit subtrees first then process all the deposit batches later.

Say if these events happen in a sequence in that interval:
- subtree1 enqueued
- subtree2 enqueued
- subtree1 dequeued and submitted to Rollup.sol
-  subtree3 enqueued
-  subtree2 dequeued and submitted to Rollup.sol

Since deposit subtrees are in a queue, the Rollup.sol always consumes from the front and new deposits are always added from the back.

ChihChengLiang (approved):
Great work! The event parsing now looks cleaner.

ts/client/services/syncer.ts
jacque006: @ChihChengLiang Would this be good to document in this code in the future to provide more context around the order?

ChihChengLiang: Sounds like a good idea

Issue #587, opened 2021-05-13
Strengthen the domain separator


ChihChengLiang:
We thank @Marik-D for finding this issue

### What's wrong

We currently have the domain separator to be the hash of the address of "Rollup.sol". It is possible the operator deploys rollup on other networks, so the new rollup has the same address and the same domain. Then a user's fund is at risk because the old txs and the old signatures can be replayed at the other networks.

https://github.com/thehubbleproject/hubble-contracts/blob/f23e37c2580ea1f40c08eafcecbba826ca041db3/contracts/rollup/Rollup.sol#L91

### How can we fix it

List some possible solutions

- At a minimum, we should add chainID to the domain.
- There's a standard EIP712, in it defines what fields we should add in a domain. We currently have `verifyingContract`, but don't have the rest of the fields.

```
string name               the user readable name of signing domain, i.e. the name of the DApp or the protocol.
string version            the current major version of the signing domain. Signatures from different versions are not compatible.
uint256 chainId           the EIP-155 chain id. The user-agent should refuse signing if it does not match the currently active chain.
address verifyingContract the address of the contract that will verify the signature. The user-agent may do contract specific phishing prevention.
bytes32 salt              an disambiguating salt for the protocol. This can be used as a domain separator of last resort.
``` 

Here's an example of Uniswap's domain separator

 https://github.com/Uniswap/uniswap-v2-periphery/blob/dda62473e2da448bc9cb8f4514dadda4aeede5f4/contracts/test/DeflatingERC20.sol#L23-L36

and its JS implementation

https://github.com/Uniswap/uniswap-v2-periphery/blob/dda62473e2da448bc9cb8f4514dadda4aeede5f4/test/shared/utilities.ts#L15-L28

PR #596, opened 2021-05-27
EIP-712 Domain

jacque006:
Resolves: https://github.com/thehubbleproject/hubble-contracts/issues/587

Ref: https://eips.ethereum.org/EIPS/eip-712

### TODO
 - [x] Add test to verify appID/domain
 - [x] Verify changes don't break BLS signing verification
 - [x] Fix broken tests

### DIFF
--- a/contracts/WithdrawManager.sol
+++ b/contracts/WithdrawManager.sol
@@ -1,15 +1,17 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.6.12;
 pragma experimental ABIEncoderV2;
-import { Types } from "./libs/Types.sol";
-import { ITokenRegistry } from "./TokenRegistry.sol";
+
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
+import { Types } from "./libs/Types.sol";
 import { Tx } from "./libs/Tx.sol";
-import { Vault } from "./Vault.sol";
 import { MerkleTree } from "./libs/MerkleTree.sol";
 import { BLS } from "./libs/BLS.sol";
 import { Bitmap } from "./libs/Bitmap.sol";
+import { IEIP712 } from "./libs/EIP712.sol";
+import { ITokenRegistry } from "./TokenRegistry.sol";
+import { Vault } from "./Vault.sol";
 
 contract WithdrawManager {
     using Tx for bytes;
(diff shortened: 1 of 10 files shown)


contracts/WithdrawManager.sol
```diff
@@ -32,7 +33,7 @@ contract WithdrawManager {
     ) public {
         tokenRegistry = _tokenRegistry;
         vault = _vault;
-        appID = keccak256(abi.encodePacked(_rollup));
+        appID = keccak256(abi.encodePacked(Rollup(_rollup).appID));
```
jacque006, line 36: We may just want to pass in the appID/domain directly here rather than coupling to the Rollup contract.

contracts/WithdrawManager.sol
```diff
@@ -91,6 +92,7 @@ contract WithdrawManager {
         (checkSuccess, callSuccess) = BLS.verifySingle(
             signature,
             pubkey,
+            // TODO What do we do here to verify EIP-712 domain?
             BLS.hashToPoint(appID, abi.encodePacked(msg.sender))
```
jacque006, line 96: @ChihChengLiang Your input on how this could/will impact BLS signing and verification is appreciated.

ChihChengLiang: The domain is appID. As long as all signing and verification use the same domain it should be fine.
EIP-712 only affects the value of the bytes32 variable.

contracts/WithdrawManager.sol
```diff
@@ -91,7 +93,10 @@ contract WithdrawManager {
         (checkSuccess, callSuccess) = BLS.verifySingle(
             signature,
             pubkey,
-            BLS.hashToPoint(appID, abi.encodePacked(msg.sender))
+            BLS.hashToPoint(
+                domain.domainSeparator(),
```
jacque006, line 97: Because the `chainid` can change (https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v3.4.0/contracts/drafts/EIP712.sol#L26-L27), we will need to check the Rollup contract every time this is run.

package.json
```diff
@@ -53,7 +53,7 @@
   "dependencies": {
     "abort-controller": "^3.0.0",
     "bn.js": "^5.2.0",
-    "ethers": "^5.1.4",
+    "ethers": "5.1.4",
```
jacque006, line 56: Per `Experimental Feature` in https://docs.ethers.io/v5/api/utils/hashing/#TypedDataEncoder , I pinned this.

contracts/WithdrawManager.sol
jacque006: Added a note for this to `IEIP712.domainSeparator()` docstring.

ChihChengLiang (approved):
Great work! The domain separator looks solid now!

test/fast/rollup.test.ts
```diff
@@ -0,0 +1,59 @@
...
+import {
+    DepositManagerFactory,
+    RollupFactory
+} from "../../types/ethers-contracts";
+import { Rollup } from "../../types/ethers-contracts/Rollup";
+
+const randomAddress = () => {
+    const hex = crypto.randomBytes(20).toString("hex");
+    return `0x${hex}`;
+};
```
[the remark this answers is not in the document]
jacque006: :+1: Will use in other PR where I move this to a util.

Issue #477, opened 2021-02-17
setRollupAddress in depositManager and Vault has no access control


ChihChengLiang:
### What's wrong

Anyone can call setRollupAddress and reset the Rollup address

### How can we fix it

- should check if the value is set before we set the value.
- Might need access control to prevent a front run.

ChihChengLiang:
Should fix with `Initializable` https://github.com/OpenZeppelin/openzeppelin-contracts/blob/440b65bf776f57cba7a1562caabd94ce41dfb12a/contracts/proxy/utils/Initializable.sol

PR #597, opened 2021-05-28
Add access control to setRollupAddress in DepositManager and Vault contracts

jacque006:
Resolves https://github.com/thehubbleproject/hubble-contracts/issues/477

### TODO
- [x] Add more access control, tests, docs for `Vault.setRollupAddress`

### DIFF
--- a/contracts/DepositManager.sol
+++ b/contracts/DepositManager.sol
@@ -1,11 +1,14 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.6.12;
 pragma experimental ABIEncoderV2;
-import { Types } from "./libs/Types.sol";
-import { ITokenRegistry } from "./TokenRegistry.sol";
+
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
+import { Initializable } from "@openzeppelin/contracts/proxy/Initializable.sol";
+import { Types } from "./libs/Types.sol";
+import { ImmutableOwnable } from "./libs/ImmutableOwnable.sol";
 import { Rollup } from "./rollup/Rollup.sol";
+import { ITokenRegistry } from "./TokenRegistry.sol";
 
 interface IDepositManager {
     event DepositQueued(uint256 pubkeyID, uint256 tokenID, uint256 l2Amount);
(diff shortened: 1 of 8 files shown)


ts/utils.ts
```diff
@@ -65,6 +65,17 @@ export function randomLeaves(num: number): string[] {
...
+/**
+ * Generates a random address. Usefully for testing when
+ * you don't need a valid address or contract.
+ *
+ * @returns Randomly generated address.
+ */
+export function randomAddress(): string {
+    const hex = crypto.randomBytes(20).toString("hex");
+    return getAddress(hex);
+}
```
jacque006, line 77: Will convert https://github.com/thehubbleproject/hubble-contracts/pull/596/files#diff-f42164895eb06a73792de9958c01fd722add2af3eee446525bc00cbfa61c7757R13-R16 to this once this PR is merged in a future PR.

ts/deploy.ts
```diff
@@ -138,7 +138,7 @@ export async function deployAll(
     );
     await waitAndRegister(rollup, "rollup", verbose);
 
-    await vault.setRollupAddress(rollup.address);
+    await waitUntilMined(vault.setRollupAddress(rollup.address));
     await waitUntilMined(depositManager.setRollupAddress(rollup.address));
```
jacque006, line 142: Would be great if we could batch calls like this but from what I've seen of typechain/ethers it's currently not possible.

ts/deploy.ts
```diff
@@ -138,7 +138,7 @@ export async function deployAll(
     );
     await waitAndRegister(rollup, "rollup", verbose);
 
-    await vault.setRollupAddress(rollup.address);
+    await waitUntilMined(vault.setRollupAddress(rollup.address));
```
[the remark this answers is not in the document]
ChihChengLiang: Right, although the order of `vault.setRollupAddress(rollup.address)` and `depositManager.setRollupAddress(rollup.address)` doesn't matter, we should wait until one is mined before we deploy the next so that if one deployment fails we have time to stop and think about it.

contracts/DepositManager.sol
```diff
@@ -119,7 +127,12 @@ contract DepositManager is DepositCore, IDepositManager {
         vault = _vault;
     }
 
-    function setRollupAddress(address _rollup) public {
+    /**
+     * @notice Sets Rollup contract address. Can only be called once by owner
+     * @dev We assume DepositManager is deployed before Rollup
+     * @param _rollup Rollup contract address
+     */
+    function setRollupAddress(address _rollup) external initializer onlyOwner {
```
ChihChengLiang, line 135: Ownable emits events and deploys public functions we don't use. What do you think we use the handcraft version? 

```
address private immutable _deployer;

constructor(){
    _deployer = msg.sender;
}

function setRollupAddress(address _rollup) external initializer {
    require(msg.sender==_deployer, "Only deployer");
    ...
}
```

ChihChengLiang (approved):
Propose an alternative to Ownable. Other parts look good to me.

Issue #519, opened 2021-03-10
Deposit feature


ChihChengLiang:
### What's wrong

Want to design the deposit feature for the client

The L1 deposit works like this. 

1. A user sends tokens to the deposit manager and tells them their public key ID. 
2. The deposit manager constructs an L2 state leaf accordingly, then accumulates the hashes of the leaf on-chain. 
3. Once enough (2**paramMaxSubtreeSize) state leaves are accumulated, the deposit manager emits an event (DepositSubTreeReady) to inform the coordinator.
4. The coordinator then finds an empty position in the state tree and places the newly created subtree there. This means the coordinator needs to create a new batch with a new state root on Rollup.sol.
5. We haven't done the deposit in an optimistic way yet. So what the coordinator needs to prove things to Rollup.sol:
    1. They prove the deposit subtree is inserted in an empty position in the state tree.
    2. To do so, they prove before the insertion, the position in the state tree is empty. So they need to provide the previous state root.
    3. The coordinator needs to make sure the previous state root is valid too. The on-chain data we have is the previous batch, the batch's commitment root, and how many commitments the batch has. So they prove the pre-state-root is indeed in a commitment, and that particular commitment is indeed from the batch and is exactly the last one.
    4. After we can trust the previous state root, the coordinator select the index of empty position in the state tree to insert that subtree. The Rollup.sol can get the correct deposit subtree root from the deposit contract, and derive the new state root.
6. After all these, the user successfully turns their L1 token to the L2 states and can spend happily hereafter. 

To handle the deposits, the client does the following things:

1. The client listens to L1 deposit events and builds subtree locally. Specifically:
    1. listen to the deposit manager for the DepositQueued event. The event arguments contain the state leaf data. The client memorizes those state leaves.
    2. listen to the deposit manager for the DepositSubTreeReady event. When a subtree is ready, a subtree root and a subtreeID are emitted. The client builds the subtree from the state leaves they memorized before.

#### Syncing node

For the syncing node, once they learned a NewBatch event emitted from Rollup.sol and the batch type is `Types.Usage.Deposit`, they need to perform the state transition to get the state view after that batch.

1. They prepare and construct the subtree locally.
2. On learning the NewBatch event, parse the calldata for `pathAtDepth` to know what position to insert the subtree.
3. Perform state transition to get the new state root.
4. (optional) validate if the commitment root matches the argument in the NewBatch event. If the state root is updated incorrectly this check would fail.

#### Packing node

[...]

PR #599, opened 2021-06-02
Deposit Packing

jacque006:
Resolves https://github.com/thehubbleproject/hubble-contracts/issues/519

### DIFF
--- a/test/client/integration.test.ts
+++ b/test/client/integration.test.ts
@@ -3,13 +3,8 @@ import { assert } from "chai";
 import { providers } from "ethers";
 import { ethers } from "hardhat";
 import { CoreAPI } from "../../ts/client/coreAPI";
-import { ConcreteBatch } from "../../ts/client/features/base";
-import { DepositCommitment } from "../../ts/client/features/deposit";
-import { GenesisCommitment } from "../../ts/client/features/genesis";
-import {
-    SimulatorPool,
-    TransferPackingCommand
-} from "../../ts/client/features/transfer";
+import { SimulatorPool, TransferPool } from "../../ts/client/features/transfer";
+import { Packer } from "../../ts/client/services/packer";
 import { SyncerService } from "../../ts/client/services/syncer";
 import { PRODUCTION_PARAMS } from "../../ts/constants";
 import { CommonToken } from "../../ts/decimal";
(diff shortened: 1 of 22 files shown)


test/client/integration.test.ts
```diff
@@ -40,12 +29,13 @@ describe("Client Integration", function() {
     it("run", async function() {
         await mcl.init();
         const [signer] = await ethers.getSigners();
-        const provider = signer.provider as providers.Provider;
+        const provider = signer.provider as providers.JsonRpcProvider;
```
jacque006, line 32: In a version of these tests I tried to get two full nodes running, during which I discovered is the only `ethers` Provider that has signers is `JsonRpcProvider`.

I can revert this back to the more general interface, but felt this was worth leaving in to call out the specific provider we use for the node.

ts/client/features/deposit.ts
```diff
@@ -184,7 +175,64 @@ export class DepositHandlingStrategy implements BatchHandlingStrategy {
...
+            subtree.states
+        );
+        await state.commit();
+
+        const context = {
+            /*
+             * TODO PR Comment Why do we need the subtreeID in this context?
+             * It is not currently being used anywhere in the client.
+             */
+            subtreeID: -1,
```
jacque006, line 227: Feedback and thoughts appreciated on this.

ts/client/features/transfer.ts
```diff
@@ -561,40 +584,25 @@ async function packBatch(
...
-                storageManager,
-                params,
-                verifier
-            );
-            commitments.push(commit);
-            txBundle.acceptedTxs.push(...acceptedTxs);
-            txBundle.failedTxs.push(...failedTxs);
-        } catch (err) {
-            console.error(err);
-            continue;
```
jacque006, line 576: Rather than suppress errors, at this point I think we should let most things bubble to the top so we can discover and fix issues in the node and tests more easily. This would just give me an error message, now it gives a full stack and stops.

ts/client/features/deposit.ts
ChihChengLiang: Good question. The subtreeID is the reliable unique identifier for the deposit subtree. We can work without subtreeID in the happy case, but when we need to debug, handle reorg, present deposits in logs or a dashboard, the subtreeID would come in handy.

(The subtreeID was proposed in #472)

ChihChengLiang (approved):
I reviewed the part that affects the protocol most and skipped all the coding style-related diff. Looks good to me.

ts/client/features/deposit.ts
jacque006: Also validate the subtreeId is valid

Issue #600, opened 2021-06-08
Add expected batchID to non-genesis Batch submissions


jacque006:
### What's wrong

When running multiple batch submissions in a single L1 transaction, if one of the batch submissions earlier or in the middle of the call fails the latter submission will still proceed, be invalid since it depended on the earlier submission's state, and be slashed.

3 Submission 1 L1 Transaction Example:
1. Submission 1 succeeds
2. Submission 2 fails
3. Submission 3 succeeds, but will be slashed as it relied on Submission 2's state changes/proof.

### How we can solve

Add an additional argument in `Rollup.sol` of `batchID/expectedBatchID` to `submitDeposits`, `submitTransfer`, `submitCreate2Transfer`, and `submitMassMigration` which is `require` checked against the current contract `nextBatchID (batchID + 1)` to ensure the batch submission will be valid.

PR #601, opened 2021-06-09
Add batchID verification to Rollup submit* functions

jacque006:
Resolves https://github.com/thehubbleproject/hubble-contracts/issues/600

TODO:
- [x] Update existing tests
- [x] Add test coverage for new paths/branches in Rollup contract
- [x] Add test coverage for new `BatchStorage` functions

### DIFF
--- a/contracts/rollup/Rollup.sol
+++ b/contracts/rollup/Rollup.sol
@@ -107,6 +107,22 @@ contract Rollup is BatchManager, EIP712, IEIP712 {
         _;
     }
 
+    /*
+     * When running multiple batch submissions in a single transaction,
+     * if one of the earlier batch submissions fails the latter submission
+     * will still proceed, be invalid since it depended on the earlier
+     * submission's state, and be slashed.
+     *
+     * To prevent this scenario, verify the expected batchID matches
+     * nextBatchID when a submission function is executed.
+     *
+     * @param batchID expected batch ID
+     */
+    modifier correctBatchID(uint256 batchID) {
+        require(batchID == nextBatchID, "batchID does not match nextBatchID");
+        _;
+    }
+
     function checkInclusion(
         bytes32 root,
         Types.CommitmentInclusionProof memory proof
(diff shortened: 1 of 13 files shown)


scripts/benchmark.ts
```diff
@@ -10,6 +10,7 @@ const txPerCommitment = 32;
 const commitmentsPerBatch = 32;
 const blockTime = 13;
 const blockGasLimit = 12500000;
+const batchID = 1;
```
jacque006, line 13: Should this be a different value?

contracts/rollup/Rollup.sol
```diff
@@ -173,6 +173,10 @@ contract Rollup is BatchManager, EIP712, IEIP712 {
             );
     }
 
+    function validateBatchID(uint256 batchID) internal view {
```
ChihChengLiang, line 176: 1. We can make `validateBatchID` a modifier so that it is consistent with other `require` checks like `onlyCoordinator` and `isNotRollingBack`.
2. Suggest documenting the reasons for this check or referring to #600, so that future auditor/reviewer can understand this better.

scripts/benchmark.ts
ChihChengLiang: batchID=1 makes sense to me since batchID 0 is the genesis.

scripts/benchmark.ts
```diff
@@ -33,6 +34,7 @@ async function main() {
 
     const tx = await batch.submit(
         constracts.rollup,
+        batchID,
         TESTING_PARAMS.STAKE_AMOUNT
```
ChihChengLiang, line 38: What do you think we add the expected batchID as a property of the batch class? So that the submit function automatically fills in that value.
It can be an undefined property by default, and the submit raise an error if it is unset. 
But feel free to ignore this comment as we are replacing the batch class used in the contract tests with the one that the client use.

contracts/rollup/Rollup.sol
```diff
@@ -279,9 +289,11 @@ contract Rollup is BatchManager, EIP712, IEIP712 {
     }
 
     function submitDeposits(
+        uint256 batchID,
         Types.CommitmentInclusionProof memory previous,
         Types.SubtreeVacancyProof memory vacant
     ) public payable onlyCoordinator isNotRollingBack {
+        validateBatchID(batchID);
         uint256 preBatchID = nextBatchID - 1;
```
ChihChengLiang, line 297: Since we have established that nextBatchID== batchID, we can use batchID here for cheaper gas.
```suggestion
        uint256 preBatchID = batchID - 1;
```

ChihChengLiang (approved):
Looks good to me. Added some comments.
Great work adding tests for the batchID check.

ChihChengLiang:
Bench shows that TPS dropped slightly from 2720.75 to 2718.43, which seems fine.

scripts/benchmark.ts
jacque006: I think it would be better to wait until we switch. I can definitely see adding the batch ID to the newer `Batch` interface, could also make the `BatchStorage` implementation simpler.

ChihChengLiang: wait until we switch sounds good

jacque006:
TPS Update: Dropped ~1 TPS (2717.35) with PR comment updates, for a total of ~3 TPS less, or a `0.1249655%` TPS drop. Think the tradeoff is worth it :) Merging.

Issue #564, opened 2021-04-23
Update @typechain/ethers-v5 to v6


jacque006:
# @typechain/ethers-v5
https://www.npmjs.com/package/@typechain/ethers-v5
Update to latest v6

ChihChengLiang:
- We could delay the upgrade of `mcl-wasm`, as hubble-contracts will not directly depend on mcl-wasm after #565.
- We shouldn't upgrade Solidity to v0.8 right now. The contracts are audited, upgrading to Solidity v0.8 will bring breaking changes and unexpected behaviors.

jacque006:
Thanks for input @ChihChengLiang , removed `mcl-wasm` in favor of https://github.com/thehubbleproject/hubble-contracts/issues/565 and moved Solidity update to a delayed section.

I will likely break more of this up next week.

jacque006:
Split this up into multiple issues so work is better broken up.

- OpenZepplin/Solidity v0.8 https://github.com/thehubbleproject/hubble-contracts/issues/566
- NodeJS v14 https://github.com/thehubbleproject/hubble-contracts/issues/567
- Prettier v2 https://github.com/thehubbleproject/hubble-contracts/issues/568

Issue #568, opened 2021-04-26
Update to prettier v2


jacque006:
https://www.npmjs.com/package/prettier
Update to latest v2

jacque006:
Consider using eslint, adding commit message linting rules.

jacque006:
Make linter follow these rules as makes sense https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md

jacque006:
Update `prettier-plugin-solidity` as well.

PR #602, opened 2021-06-11
Update typechain, other minor deps

jacque006:
Resolves https://github.com/thehubbleproject/hubble-contracts/issues/564


package.json
```diff
@@ -33,28 +33,28 @@
...
     "minimist": "^1.2.5",
     "prettier": "^1.19.1",
-    "prettier-plugin-solidity": "1.0.0-beta.10",
-    "solhint": "^3.3.4",
-    "ts-node": "^9.1.1",
-    "typechain": "^2.0.1",
-    "typescript": "^4.2.4"
+    "prettier-plugin-solidity": "1.0.0-beta.11",
+    "solhint": "^3.3.6",
+    "ts-node": "^10.0.0",
```
jacque006, line 49: No breaking changes on our end from what I can see.

package.json
```diff
@@ -33,28 +33,28 @@
...
-    "hardhat": "^2.2.1",
+    "hardhat": "^2.3.2",
     "minimist": "^1.2.5",
     "prettier": "^1.19.1",
-    "prettier-plugin-solidity": "1.0.0-beta.10",
-    "solhint": "^3.3.4",
-    "ts-node": "^9.1.1",
-    "typechain": "^2.0.1",
-    "typescript": "^4.2.4"
+    "prettier-plugin-solidity": "1.0.0-beta.11",
```
jacque006, line 47: We will not be able to go higher on this version until https://github.com/thehubbleproject/hubble-contracts/issues/568

test/fast/accountTree.test.ts
```diff
@@ -1,11 +1,18 @@
...
 import { Tree, Hasher } from "../../ts/tree";
 import { ethers } from "hardhat";
 import { assert } from "chai";
 import { randHex, randomLeaves } from "../../ts/utils";
 import { ZERO_BYTES32 } from "../../ts/constants";
 
+type UpdateBatchLeafs = Parameters<TestAccountTree["updateBatch"]>[0];
+type CheckInclusionTreeWitness = Parameters<
+    TestAccountTree["checkInclusion"]
+>[2];
```
jacque006, line 14: This and the usage points below are the best way I could determine to allow compatibility with Solidity tuple types now being generated in TS land (uint156[2] => [BigNumberish, BigNumberish]).

Would be nice if we could verify in Node.js before making txn calls that the length of arrays is valid, but for now will have to rely on the txn failing.

ts/mcl.ts
```diff
@@ -14,8 +14,8 @@ export type MessagePoint = mclG1;
 export type Signature = mclG1;
 export type PublicKey = mclG2;
 
-export type solG1 = [string, string];
-export type solG2 = [string, string, string, string];
+export type solG1 = [BigNumberish, BigNumberish];
+export type solG2 = [BigNumberish, BigNumberish, BigNumberish, BigNumberish];
```
jacque006, line 18: Based on my understanding of `mcl` and how these types are used in TS/JS code and at the contract level, this seems like a more appropriate type.

ChihChengLiang (approved):
Looks good to me 👍

PR #605, opened 2021-06-14
Create Leaf, pubkeyLeaf and stateLeaf

kautukkundan:


ts/tree/leaves/Leaf.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    get key() {
+        return this.getKey();
+    }
+
+    serialize(): string {
+        return this.item.encode();
+    }
+
+    abstract deserialize(bytes: string): Item;
```
kautukkundan, line 39: since both Pubkey and State has a `fromEncoded` method, should we directly implement this method here or keep it abstract and implement later?

jacque006: I would just get rid of it here and it's implementations as well since we are using the static factory method with db.

kautukkundan:
- remove deserialize method

Issue #557, opened 2021-04-22
Add tokenID to feeReceiverID mapping config


ChihChengLiang:
### What's wrong

To serve tokenA, the coordinator needs to own a state leaf of tokenA to collects the fees.
When running the coordinator client, the client should be configured with a mapping of what tokenID to what stateID.

### How can we fix it?

Create an interface (Feel free to rename)

```
interface FeeReceiverConfig {
   getFeeReceiver(tokenID: number): number
}
```
Which belongs to ClientConfigs

```
interface ClientConfigs {
    feeReceiverConfig: FeeReceiverConfig
    ...
}
```

Create a config that can be loaded by the nodes. (We could defer this if we want to hardcode everything)

```
{
  #tokenID: stateID
  0: 1,
  1: 100,
}
```

jacque006:
Note, more general configuration is also likely needed for RPC port, ETH client ip/port/url/connection, and other options. This will also be needed for any assets we release (Docker, npm, etc.)

jacque006:
As part of this work, we should also validate that the config is correct and throw a very specific error is a config option is missing or malformed.

jacque006:
After completing this and confirming Docker based image can run successfully, update DockerHub Overview at https://hub.docker.com/r/thehubbleproject/node (and potentially automate syncing with GitHub README.md) with instructions on how to run container.

Issue #408, opened 2021-01-08
Adding a coordinator registry that has their IP address or API endpoint


ChihChengLiang:
### What's wrong

User needs to know where to send their layer2 transactions to the coordinator.

### How can we fix it

Create a contract that lets coordinator register and update their IP addresses.

jacque006:
Consider updating POST `/tx` route in client with information about who the active proposer is https://github.com/thehubbleproject/hubble-contracts/pull/606#discussion_r660318016

PR #606, opened 2021-06-15
Add config file, multi-token config support, node mode updates

jacque006:
Resolves https://github.com/thehubbleproject/hubble-contracts/issues/557

### TODO
- [x] Make sure client (node) spins up without config file.
- [x] Update SETUP.md, docker docs


ts/client/services/rpc.ts
```diff
@@ -19,70 +15,54 @@ const tx = {
...
+            }
+        );
+        fastify.post<{ Body: { bytes: string } }>(
+            "/tx",
+            tx,
+            async (request, reply) => {
+                // TODO There should be a better way we can reject
+                // accepting transactions when not a proposer.
+                // Also, do we want to accept transactions when we are not
+                // the active proposer?
```
jacque006, line 38: Thoughts?

ts/genesis.ts
```diff
@@ -35,17 +41,24 @@ export class Genesis {
...
 
-    static fromConfig(path: string) {
-        const genesis = fs.readFileSync(path).toString();
-        const { parameters, addresses, auxiliary } = JSON.parse(genesis);
+    public static async fromConfig(
+        path: string = "./genesis.json"
+    ): Promise<Genesis> {
+        const genesisPath = resolve(path);
+        console.info(`loading genesis from ${genesisPath}`);
+        const genesisJSONStr = await readFileAsync(genesisPath, {
```
jacque006, line 49: non-blocking NodeJS i/o APIs are preferred over `somethingSync` so we don't block the event loop.

ts/client/services/rpc.ts
ChihChengLiang: Good question. Haven't given this a good thought yet. But my first thoughts are

- A proposer should not accept tx when not active.
- Ideally, the proposer can reply "hey the current active proposer endpoint is https://theotherprovider.com". But since we are talking to wallets, `reply.status(409).send("not a proposer")` seems fine. The application dev should handle this gracefully.

ChihChengLiang (approved):
Great works! Looks good to me

ts/client/services/rpc.ts
jacque006: Removed TODO, added note to https://github.com/thehubbleproject/hubble-contracts/issues/408#issuecomment-870768409

kautukkundan (approved):
looks good 👍

it gave me an opportunity to checkout `TransferPackingCommand` and `DepositPackingCommand`. Will dig deeper.

PR #609, opened 2021-06-20
Create DB tree and DB engine

kautukkundan:
- created Tree interface
- memoryTree implements Tree 
- replaced all instances of Tree with MemoryTree
- created dbTree extends memoryTree
- created dbEngine extends memoryEngine 
- replaced memoryTree with dbTree in accountTree
- changed memoryEngine to dbEngine in factory

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ cache/
 artifacts/
 build/
 docker/testData
+leveldb/
--- a/ts/client/database/databaseEngine.ts
+++ b/ts/client/database/databaseEngine.ts
@@ -1,0 +1,12 @@
+import { Hashable } from "../../interfaces";
+import { Leaf, LeafFactoryFunc } from "../../tree/leaves/Leaf";
+import { MemoryEngine } from "../storageEngine/memoryEngine";
+
+export class DatabaseEngine<Item extends Hashable> extends MemoryEngine<Item> {
+    private readonly factory: LeafFactoryFunc<Leaf<Item>>;
+
+    constructor(depth: number, factory: LeafFactoryFunc<Leaf<Item>>) {
+        super(depth);
+        this.factory = factory;
+    }
+}
--- a/ts/client/features/base.ts
+++ b/ts/client/features/base.ts
@@ -1,6 +1,6 @@
 import { BytesLike } from "ethers";
 import { solidityKeccak256 } from "ethers/lib/utils";
-import { Tree } from "../../tree";
+import { MemoryTree } from "../../tree/memoryTree";
 import { prettyHex } from "../../utils";
 import {
     Batch,
@@ -31,10 +31,10 @@ export abstract class BaseCommitment implements Commitment {
 }
 
 export class ConcreteBatch<T extends Commitment> implements Batch {
-    private tree: Tree;
+    private tree: MemoryTree;
     constructor(public readonly commitments: T[]) {
         if (commitments.length === 0) throw new Error("no commitment");
-        this.tree = Tree.merklize(commitments.map(c => c.hash()));
+        this.tree = MemoryTree.merklize(commitments.map(c => c.hash()));
     }
 
     get postStateRoot(): string {
--- a/ts/commitments.ts
+++ b/ts/commitments.ts
@@ -6,7 +6,7 @@ import { Usage, Wei } from "./interfaces";
 import { solG1 } from "./mcl";
 import { State } from "./state";
 import { MigrationTree, StateProvider } from "./stateTree";
-import { Tree } from "./tree";
+import { MemoryTree } from "./tree/memoryTree";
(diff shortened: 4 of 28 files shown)


.gitignore
```diff
@@ -6,3 +6,4 @@ cache/
 artifacts/
 build/
 docker/testData
+leveldb/
```
jacque006, line 9: #nitpick newline after

ts/client/database/databaseEngine.ts
```diff
@@ -0,0 +1,129 @@
...
+import { Hashable, Vacant } from "../../interfaces";
+import { Hasher } from "../../tree";
+import { DBTree } from "../../tree/dbTree";
+import { Leaf, LeafFactoryFunc } from "../../tree/leaves/Leaf";
+import { StorageEngine, WithWitness } from "../storageEngine/interfaces";
+
+export interface Entry<Item> {
+    itemID: number;
+    item: Item;
+}
```
jacque006, line 12: This should be moved to `storageEngine/interfaces` and used here and in `MemoryStorageEngine`

ts/client/database/databaseEngine.ts
```diff
@@ -0,0 +1,129 @@
...
+
+export class DatabaseEngine<Item extends Hashable>
+    implements StorageEngine<Item> {
+    private tree: DBTree<Leaf<Item>>;
+    private items: { [key: number]: Item } = {};
+    private cache: { [key: number]: Item } = {};
+    private journal: Entry<Item>[] = [];
```
jacque006, line 19: Per comments from @ChihChengLiang , we don't really need journal or cache and its corresponding calls to flush such as `trueUpdate`. We can remove this and make the `commit` function a no-op and write directly to db.

ts/client/database/databaseEngine.ts
```diff
@@ -0,0 +1,129 @@
...
+
+export interface Entry<Item> {
+    itemID: number;
+    item: Item;
+}
+
+export class DatabaseEngine<Item extends Hashable>
+    implements StorageEngine<Item> {
+    private tree: DBTree<Leaf<Item>>;
+    private items: { [key: number]: Item } = {};
```
jacque006, line 17: This can be replaced with LevelDB storage

ts/client/features/base.ts
```diff
@@ -31,10 +31,10 @@ export abstract class BaseCommitment implements Commitment {
 }
 
 export class ConcreteBatch<T extends Commitment> implements Batch {
-    private tree: Tree;
+    private tree: MemoryTree;
```
jacque006, line 34: This can be the `Tree` interface

ts/commitments.ts
```diff
@@ -237,9 +237,9 @@ export class Create2TransferCommitment extends TransferCommitment {
 }
 
 export class Batch {
-    private tree: Tree;
+    private tree: MemoryTree;
```
jacque006, line 240: `Tree` interface

ts/stateTree.ts
```diff
@@ -92,10 +93,10 @@ export class StateTree implements StateProvider {
     public static new(stateDepth: number) {
         return new StateTree(stateDepth);
     }
-    private stateTree: Tree;
+    private stateTree: MemoryTree;
```
jacque006, line 96: Unsure if this should be a memory or database tree, @ChihChengLiang can you provide insight?

jacque006 (commented):
Good start to getting this setup. In addiction to comments, I think there is file restructuring we can do to organize some of this new code a bit better. Will follow up with off-review to discuss.

ts/stateTree.ts
ChihChengLiang: As all classes in the stateTree.ts here are currently used in the tests, using MemoryTree should be fine.

Eventually, we want to replace all classes in the stateTree.ts with the classes used by the client.

ts/client/features/base.ts
kautukkundan: It uses `.merklize` static method from the MemoryTree class. Thought it would be better to give keep the type of tree as MemoryTree due to that method. What do you think?

ts/commitments.ts
kautukkundan: same as prev

ts/client/features/base.ts
jacque006: We can make the `merklize` function return the `Tree` interface and still use that here. Fine as is, can do on next PR.

jacque006 (approved):

jacque006:
Last thing to do before merging is to cleanup commits.

`git rebase -i HEAD~9` and use the interactive git workflow to squash down to 1 commit with details for all this work.
Then `git push --force-with-lease origin db-memory-tree` to push changes up overriding commits that are already here.

Issue #612, opened 2021-06-22
Bug: Unhandled Promise rejection when bidding on proposer slot


jacque006:
```
block 1344      Slot progress   9/10
block 1345      Slot progress   0/10
New slot 128
Auctioning slot coordinator 0x0000000000000000000000000000000000000000 0.0 ETH
Bid 0.000000000000000001 ETH
(node:15276) UnhandledPromiseRejectionWarning: Error: invalid BigNumber value (argument="value", value=null, code=INVALID_ARGUMENT, version=bignumber/5.3.0)
    at Logger.makeError (/root/hubble/hubble-contracts/node_modules/@ethersproject/logger/src.ts/index.ts:213:28)
    at Logger.throwError (/root/hubble/hubble-contracts/node_modules/@ethersproject/logger/src.ts/index.ts:225:20)
    at Logger.throwArgumentError (/root/hubble/hubble-contracts/node_modules/@ethersproject/logger/src.ts/index.ts:229:21)
    at Function.BigNumber.from (/root/hubble/hubble-contracts/node_modules/@ethersproject/bignumber/src.ts/bignumber.ts:291:23)
    at Formatter.bigNumber (/root/hubble/hubble-contracts/node_modules/@ethersproject/providers/src.ts/formatter.ts:183:26)
    at Function.Formatter.check (/root/hubble/hubble-contracts/node_modules/@ethersproject/providers/src.ts/formatter.ts:428:42)
    at Formatter.transactionResponse (/root/hubble/hubble-contracts/node_modules/@ethersproject/providers/src.ts/formatter.ts:324:55)
    at JsonRpcProvider.<anonymous> (/root/hubble/hubble-contracts/node_modules/@ethersproject/providers/src.ts/base-provider.ts:1412:39)
    at step (/root/hubble/hubble-contracts/node_modules/@ethersproject/providers/lib/base-provider.js:48:23)
    at Object.next (/root/hubble/hubble-contracts/node_modules/@ethersproject/providers/lib/base-provider.js:29:53)
(node:15276) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 19)
block 1346      Slot progress   1/10
```

[...]

PR #618, opened 2021-06-28
Update dependencies

jacque006:
Resolves https://github.com/thehubbleproject/hubble-contracts/issues/612

Update ethers.js for EIP-1559 support.
Update other non-breaking dependencies to latest versions.


test/integration.test.ts
```diff
@@ -354,7 +354,9 @@ describe("Integration Test", function() {
         for (const batchID of stakedBatchIDs) {
             const preBalance = await coordinator.getBalance();
             const tx = await rollup.connect(coordinator).withdrawStake(batchID);
-            const txFee = (await tx.wait()).gasUsed.mul(tx.gasPrice);
+            const txReceipt = await tx.wait();
+            assert.isDefined(tx.gasPrice);
+            const txFee = txReceipt.gasUsed.mul(tx.gasPrice ?? -1);
```
jacque006, line 359: Is there a better default value we can use asides from `-1`? We should never hit this case with the `assert.isDefined` above but it's needed for TypeScript.

ChihChengLiang: I used to write `if (tx.gasPrice === undefined) throw Error()` and then TypeScript doesn't complain the next line.
Writing `-1` sounds fine to me. If we really worried we can comment something like "write -1 here to make ts happy".

ChihChengLiang (approved):
LGTM

kautukkundan (approved):
LGTM 👍

Issue #567, opened 2021-04-26
Update NodeJS v14


jacque006:
NodeJS

Update to latest LTS v14

- Add an [engines](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#engines) entry to our `package.json` for LTS versions v10 - v14
- Add build matrix entries in Github Actions for v12, v14 that we only run on merges to master/releases. Run these only for client tests, use v14 for other tests.

jacque006:
`npm@7` introduced a new lockfile format https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json#lockfileversion . As part of moving to a new node version, we should move to npm v7 and enforce that version or greater in our engines entry.

jacque006:
In version matrix, consider adding v16.x as well since it will LTS in October

jacque006:
Remove `del` dep https://github.com/thehubbleproject/hubble-contracts/pull/619#discussion_r668288517

jacque006:
Update `target` and `lib` entries in `tsconfig.json` to match proper ES support level for node version.

PR #619, opened 2021-07-05
complete DB integration with AccountTree

### LINKED ISSUE #614: Implement DB Tree and DB Engine Logic
Currently the DB Tree and DB Engine are simply extensions of Memory Tree and Memory engine respective. Need to Implement separate logic for each to make it compatible with levelDB 

https://github.com/thehubbleproject/hubble-contracts/blob/master/ts/tree/dbTree.ts

https://github.com/thehubbleproject/hubble-contracts/blob/master/ts/client/database/databaseEngine.ts

kautukkundan:
closes #614

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "@types/subleveldown": "^4.1.0",
     "chai": "^4.3.4",
     "chai-as-promised": "^7.1.1",
+    "del": "^6.0.0",
     "hardhat": "^2.4.1",
     "prettier": "^1.19.1",
     "prettier-plugin-solidity": "1.0.0-beta.11",
--- a/test/client/integration.test.ts
+++ b/test/client/integration.test.ts
@@ -13,6 +13,7 @@ import { deployKeyless } from "../../ts/deployment/deploy";
 import { Group, storageManagerFactory } from "../../ts/factory";
 import { Genesis } from "../../ts/genesis";
 import * as mcl from "../../ts/mcl";
+import del from "del";
 
 /**
  * This integration test ensures that
@@ -26,6 +27,14 @@ import * as mcl from "../../ts/mcl";
  * and reach a correct end state.
  */
 describe("Client Integration", function() {
+    before(async function() {
+        await del("./leveldb/*");
+    });
+
+    after(async function() {
+        await del("./leveldb/*");
+    });
+
     it("run", async function() {
         await mcl.init();
         const [signer] = await ethers.getSigners();
--- a/ts/client/database/databaseEngine.ts
+++ b/ts/client/database/databaseEngine.ts
@@ -1,12 +1,103 @@
-import { Hashable } from "../../interfaces";
-import { Leaf, LeafFactoryFunc } from "../../tree/leaves/Leaf";
-import { MemoryEngine } from "../storageEngine/memoryEngine";
+import { ZERO_BYTES32 } from "../../constants";
+import { TreeAtLevelIsFull } from "../../exceptions";
+import { Hashable, Vacant } from "../../interfaces";
+import { Hasher } from "../../tree";
+import { DBTree } from "../../tree/dbTree";
+import { LeafFactoryFunc } from "../../tree/leaves/Leaf";
+import { StorageEngine, WithWitness } from "../storageEngine/interfaces";
 
-export class DatabaseEngine<Item extends Hashable> extends MemoryEngine<Item> {
-    private readonly factory: LeafFactoryFunc<Leaf<Item>>;
+export class DatabaseEngine<Item extends Hashable>
+    implements StorageEngine<Item> {
+    private tree: DBTree;
+    public readonly leafFactory: LeafFactoryFunc<Item>;
 
-    constructor(depth: number, factory: LeafFactoryFunc<Leaf<Item>>) {
-        super(depth);
-        this.factory = factory;
+    constructor(depth: number, factory: LeafFactoryFunc<Item>) {
+        this.tree = DBTree.new(
+            depth,
+            factory.name,
+            Hasher.new("bytes", ZERO_BYTES32)
+        );
+        this.leafFactory = factory;
+    }
+
+    private checkSize(itemID: number) {
+        if (itemID >= this.tree.setSize)
+            throw new Error(
+                `Want itemID ${itemID} but the tree has only ${this.tree.setSize} leaves`
+            );
+    }
+
+    public get root() {
+        return this.tree.root;
+    }
+
+    public async get(itemID: number): Promise<Item> {
+        this.checkSize(itemID);
+        return (await this.leafFactory.fromDB(itemID)).item;
+    }
+
+    public async getWithWitness(itemID: number): Promise<WithWitness<Item>> {
+        const item = await this.get(itemID);
+        const witness = (await this.tree.witness(itemID)).nodes;
+        return { item, witness };
+    }
+
+    public async update(itemID: number, item: Item) {
+        this.checkSize(itemID);
+        await this.trueUpdate(itemID, item);
+    }
+
+    private async trueUpdate(itemID: number, item: Item) {
+        await this.leafFactory.newLeaf(item, itemID).toDB();
+        await this.tree.updateSingle(itemID, item.hash());
+    }
+
+    public async create(itemID: number, item: Item) {
+        const itemFound = await this.get(itemID);
+        if (itemFound) throw new Error(`Already exists  itemID: ${itemID}`);
+        await this.update(itemID, item);
+    }
+
+    public getCheckpoint(): number {
+        return 0;
+    }
+
+    public revert(checkpoint: number = 0) {
+        // no-op
+    }
+
+    public async commit() {
+        // no-op
+    }
+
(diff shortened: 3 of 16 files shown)


ts/tree/dbTree.ts
```diff
@@ -1,17 +1,210 @@
...
+    constructor(depth: number, nodeType: string, hasher: Hasher) {
+        this.depth = depth;
+        this.hasher = hasher;
+        this.zeros = this.hasher.zeros(depth);
+        this.setSize = 2 ** this.depth;
+        this.nodeType = nodeType;
+    }
+
+    public async root(): Promise<Node> {
+        return await this.getNode(0, 0);
```
ChihChengLiang, line 33: We can add a `_root` variable to cache the root to make the `root` function stay sync. So that we don't have to change all over the codebase to use async function.

The `_root` should be updated in the `ascend` function.

kautukkundan: set the default value zeros[0] and create a function to set the root during restart

ts/tree/tree.ts
```diff
@@ -29,3 +30,23 @@ export interface Tree {
     updateBatch(offset: number, leaves: Array<Node>): void;
     isZero(level: number, leafIndex: number): boolean;
 }
+
+export interface AsyncTree {
```
kautukkundan, line 34: change memory tree to use promises but use same interface for both memory and db tree

kautukkundan: keeping independent Async Tree to not mess with Memory Tree

package.json
```diff
@@ -45,6 +45,7 @@
     "@types/subleveldown": "^4.1.0",
     "chai": "^4.3.4",
     "chai-as-promised": "^7.1.1",
+    "del": "^6.0.0",
```
kautukkundan, line 48: had to install this module because `rmdirAsync` from fs module wasn't compatible with node v10 for recursively deleting a folder.

ts/tree/dbTree.ts
```diff
@@ -1,17 +1,218 @@
...
+import { ItemNode } from "./leaves/Node";
+import { AsyncTree, Data, Witness } from "./tree";
+
+export class DBTree implements AsyncTree {
+    public readonly zeros: Array<Node>;
+    public readonly depth: number;
+    public readonly setSize: number;
+    public readonly hasher: Hasher;
+    public readonly nodeType: string;
+    public _root: Node;
```
kautukkundan, line 19: using this variable as a cache for root which gets updated after every "ascend"

ts/tree/dbTree.ts
```diff
@@ -1,17 +1,218 @@
...
+                len += 1;
+            }
+            for (let node = offset; node < offset + len; node += 2) {
+                await this.updateCouple(level, node);
+            }
+            offset >>= 1;
+            len >>= 1;
+        }
+
+        await this.updateRoot();
```
kautukkundan, line 195: updates the cached root

package.json
jacque006: Can we remove this and use `rmdirAsync` once https://github.com/thehubbleproject/hubble-contracts/issues/567 is completed? If so I would add a comment to that issue to that effect.

test/client/integration.test.ts
```diff
@@ -26,6 +27,14 @@ import * as mcl from "../../ts/mcl";
  * and reach a correct end state.
  */
 describe("Client Integration", function() {
+    before(async () => {
+        await del("./leveldb");
+    });
+
+    after(async () => {
+        await del("./leveldb");
+    });
```
jacque006, line 36:
```suggestion
    before(async function () {
        await del("./leveldb");
    });

    after(async function () {
        await del("./leveldb");
    });
```
See https://mochajs.org/#arrow-functions . Generally in our test suite it will not matter, but good to follow Mocha's recommendations. Arrow funcs are also generally fine as callbacks for `descibe` blocks.

test/client/integration.test.ts
```diff
@@ -26,6 +27,14 @@ import * as mcl from "../../ts/mcl";
  * and reach a correct end state.
  */
 describe("Client Integration", function() {
+    before(async () => {
+        await del("./leveldb");
```
jacque006, line 31: We should make the leveldb path configurable in future work, out of scope for this PR.

ts/client/database/databaseEngine.ts
```diff
@@ -1,12 +1,107 @@
...
+        return this.tree.root;
+    }
+
+    public async get(itemID: number): Promise<Item> {
+        this.checkSize(itemID);
+        try {
+            return (await this.leafFactory.fromDB(itemID)).item;
+        } catch (error) {
+            throw new Error(`Item not exists  itemID: ${itemID}`);
+        }
```
jacque006, line 40: Is this `try/catch` needed, or will the underlying leveldb `get` in `leafFactory.fromDB` throw? If `get` can return an undefined value, this would be a better not found approach:
```suggestion
        const leaf  = await this.leafFactory.fromDB(itemID);
        if (!leaf) {
            throw new Error(`Item not exists  itemID: ${itemID}`);
        }
        return leaf.item;
```

ts/client/database/databaseEngine.ts
```diff
@@ -1,12 +1,107 @@
...
+        return 0;
+    }
+
+    public revert(checkpoint: number = 0) {
+        // no-op
+    }
+
+    public async commit() {
+        // no-op
+    }
```
jacque006, line 75: :+1:

ts/tree/dbTree.ts
```diff
@@ -1,17 +1,218 @@
...
+    get root(): Node {
+        return this._root;
+    }
+
+    public async getNode(level: number, index: number): Promise<Node> {
+        try {
+            return await ItemNode.fromDB(this.nodeType, level, index);
+        } catch (e) {
+            return this.zeros[level];
+        }
```
jacque006, line 43: This could suppress other errors asides from the item not being found, can we do an undefined check on `ItemNode.fromDB` and throw instead?

ts/tree/leaves/Node.ts
```diff
@@ -0,0 +1,38 @@
...
+): string => `${name}${depth}${index}`;
+
+export class ItemNode {
+    static async fromDB(
+        name: string,
+        depth: number,
+        index: number
+    ): Promise<Node> {
+        const key = getNodeKey(name, depth, index);
+        return await nodeDB.get(key);
```
jacque006, line 17: `await`ing is not needed in async chains when returning values, see this discussion https://github.com/thehubbleproject/hubble-contracts/pull/556#discussion_r618636732
```suggestion
        return nodeDB.get(key);
```

jacque006 (changes requested):
Overall looks good. Some comments mainly around exception throwing/handling.

Would be great if we could get a test file that covers the async tree implementation for both memory and db, there's some starting logic in `stateEngine.test.ts` for the memory engine/tree. Would ensure we have edge cases and behavior properly covered, and that we can 1-1 swap tree/storage engine implementations. Think it would be fine to do that as follow up to this PR.

ts/tree/dbTree.ts
```diff
@@ -1,17 +1,218 @@
...
+        return new DBTree(depth, nodeType, hasher || Hasher.new());
+    }
+
+    constructor(depth: number, nodeType: string, hasher: Hasher) {
+        this.depth = depth;
+        this.hasher = hasher;
+        this.zeros = this.hasher.zeros(depth);
+        this.setSize = 2 ** this.depth;
+        this.nodeType = nodeType;
+        this._root = this.zeros[0];
```
ChihChengLiang, line 31: I just comment to highlight issues here. We can keep it as is in this PR and fix the problem in the next PR.

This works when we have an empty tree. But we will have a problem once we have new leaves in the tree and have a non-empty tree root in DB.

When we restart the client and read the `root()` from the DBTree, we expect to read the non-empty tree root in DB. But here the cached root still returns the empty tree root.

We will need 2 steps. One reads the cache from DB and a step dumps the cache to DB.

package.json
kautukkundan: Yeah sure we can

ts/client/database/databaseEngine.ts
kautukkundan: yeah, get throws an error if the key is not present in the db. 
`NotFoundError: Key not found in database`

test/client/integration.test.ts
kautukkundan: ah, wasn't aware of this. will make changes.

ts/tree/leaves/Node.ts
kautukkundan: gotcha, thanks for the detailed explanation 😄

ts/tree/dbTree.ts
kautukkundan: absolutely! I had planned on discussing this on the call. We can come up with a `restoreFromDB` function which can check if a tree already exists and if yes, it can load the tree root. Was a bit confused though as to where we will be calling that.

ts/tree/dbTree.ts
kautukkundan: check if key exists

jacque006 (approved):

PR #620, opened 2021-07-07
changed tree to use promises

kautukkundan:
using async/await pattern everywhere to facilitate migrating to db tree

### DIFF
--- a/ts/commitments.ts
+++ b/ts/commitments.ts
@@ -42,7 +42,7 @@ abstract class Commitment {
         );
     }
     abstract toSolStruct(): SolStruct;
-    abstract toBatch(): Batch;
+    abstract toBatch(): Promise<Batch>;
     public toCompressedStruct(): CompressedStruct {
         return {
             stateRoot: this.stateRoot,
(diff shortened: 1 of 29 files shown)


ts/commitments.ts
```diff
@@ -225,58 +226,70 @@ export class MassMigrationCommitment extends Commitment {
...
+    public async toBatch(): Promise<Create2TransferBatch> {
+        return await Create2TransferBatch.new([this]);
     }
 }
 
 export class Batch {
-    private tree: MemoryTree;
-    constructor(public readonly commitments: Commitment[]) {
-        this.tree = MemoryTree.merklize(commitments.map(c => c.hash()));
+    public tree: MemoryTree = MemoryTree.new(1);
```
kautukkundan, line 241: not proud of this workaround, please suggest a better way to do this

ts/commitments.ts
```diff
@@ -225,58 +226,70 @@ export class MassMigrationCommitment extends Commitment {
...
+    constructor(public readonly commitments: Commitment[]) {}
+
+    static async new(commitments: Commitment[]) {
+        const batch = new Batch(commitments);
+        batch.tree = await MemoryTree.merklize(commitments.map(c => c.hash()));
```
kautukkundan, line 246: creates a Batch object with default MemoryTree and then sets a new Tree to it after init

kautukkundan:
closing wrt our latest discussion

Issue #476, opened 2021-02-14
Scripts to add 4 tokens and fee receivers for dev case


ChihChengLiang:

jacque006:
@ChihChengLiang Can this be closed in favor of https://github.com/thehubbleproject/hubble-contracts/issues/557? Or is this for setting up the contract where those are stored?

ChihChengLiang:
Yes, this is for some more stuff.

1. setting up the contracts to register 4 tokens in the tokenRegistry
2. revamping the txpool to support multiple token.
3. (optional) figure out how to compare token fees between different tokens. Need a price feed for this. We could defer this by using a more naive strategy like serving each token one by one.

jacque006:
Multi token tx pool broken off into https://github.com/thehubbleproject/hubble-contracts/issues/626

PR #621, opened 2021-07-08
Multi Token Pool

jacque006:
Resolves https://github.com/thehubbleproject/hubble-contracts/issues/626

### TODO
- [x] Unit test `MultiTokenPool`
- [x] Add multiple tokens to client integration tests
~Add script to generate and deposit (create state) for fee receivers~ https://github.com/thehubbleproject/hubble-contracts/issues/476
- [x] Double check `number` -> `BigNumber` refactor points.


scripts/deploy.ts
```diff
@@ -80,10 +80,11 @@ async function main() {
             `Registering ${numPubkeys} pubkeys. Custom mnemonic: ${!!pubkeyMnemonic}`
         );
         const group = Group.new({
-            n: pubkeyMnemonic,
+            n: parseInt(numPubkeys),
```
jacque006, line 83: Pretty sure this derp was me, should work properly now.

ts/client/features/deposit.ts
```diff
@@ -116,7 +116,7 @@ export async function handleNewBatch(
     const receipt = await event.getTransactionReceipt();
     const logs = receipt.logs.map(log => rollup.interface.parseLog(log));
     const depositsFinalisedLog = logs.filter(
-        log => log.signature == "DepositsFinalised(uint256,bytes32,uint256)"
+        log => log.signature === "DepositsFinalised(uint256,bytes32,uint256)"
```
jacque006, line 119: `===` > `==` https://stackoverflow.com/a/3586781

ts/client/storageEngine/interfaces.ts
```diff
@@ -10,6 +10,8 @@ export interface Entry<Item> {
     item: Item;
 }
 
+// TODO Update this interface and implementations to
+// use BigNumber for itemID.
```
jacque006, line 14: Avoided this refactor as this PR is already large and didn't want to bork @kautukkundan 's work :) `BigNumberish` should be fine as well as long as the storage engine and tree both store in a large value format under the covers.

ts/mcl.ts
```diff
@@ -176,7 +176,7 @@ export function parseG2(solG2: solG2): mclG2 {
 }
 
 function dump(sol: solG1 | solG2): string {
-    return `0x${sol.map(n => n.toString().slice(2)).join()}`;
+    return `0x${sol.map(n => n.toString().slice(2)).join("")}`;
```
jacque006, line 179: This is also my derp. `join` by default adds a comma separator which the previous implementation did not.

jacque006:
@kautukkundan @ChihChengLiang your :eyes: would be especially appreciated on `number` -> `BigNumber` refactor points. I avoided updating code in `commitments.ts` and `StorageEngine`, mainly focused on values in `ts/features/transfers`.

contracts/test/CustomToken.sol
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * @dev Contract for use testing multi-token scenarios.
+ */
+contract CustomToken is ERC20, Ownable {
+    constructor(string memory name, string memory symbol)
+        public
+        ERC20(name, symbol)
+    {
+        _mint(msg.sender, 1000000000000000000000000000);
```
kautukkundan, line 15: can use exponential notation `1e27` for simplicity

contracts/test/ExampleToken.sol
```diff
@@ -6,7 +6,8 @@ import "@openzeppelin/contracts/access/Ownable.sol";
 
 contract ExampleToken is ERC20, Ownable {
     /**
-     * @dev assign totalSupply to account creating this contract */
+     * @dev assign totalSupply to account creating this contract
+     */
     constructor() public ERC20("Example", "EMP") {
         _mint(msg.sender, 1000000000000000000000000000);
```
kautukkundan, line 12: same exponential notation can be used here

test/client/integration.test.ts
```diff
@@ -60,49 +61,69 @@ describe("Client Integration", function() {
 
         const contracts = await deployAll(signer, parameters);
 
+        // Setup and register custom tokens
+        const customTokens = await Promise.all([
+            new CustomToken__factory(signer).deploy("Hubble", "HUB"),
+            new CustomToken__factory(signer).deploy("Telescope", "TLSC")
```
kautukkundan, line 67: Telescope token ❤️

test/client/transaction.test.ts
```diff
@@ -13,6 +14,22 @@ import {
 
 chai.use(chaiAsPromised);
 
+const txFactory = (
```
kautukkundan, line 17: 👍🏼

ts/client/storageEngine/interfaces.ts
kautukkundan: oh makes sense, can discuss this

ts/client/config.ts
```diff
@@ -1,12 +1,13 @@
+import { BigNumberish } from "ethers";
 import { readFile } from "fs";
 import { resolve } from "path";
 import { promisify } from "util";
 
 const readFileAsync = promisify(readFile);
 
 export type FeeReceivers = Array<{
```
ChihChengLiang, line 8: What do you think we make FeeReceivers a class, so we can build the get random token getter or iterators from the object method?

test/fast/txPool.test.ts
```diff
@@ -0,0 +1,278 @@
...
+    tokenID: BigNumber;
+    feeReceiverID: BigNumber;
+    sumFees: BigNumber;
+}): { tokenID: number; feeReceiverID: number; sumFees: number } => ({
+    tokenID: tokenID.toNumber(),
+    feeReceiverID: feeReceiverID.toNumber(),
+    sumFees: sumFees.toNumber()
+});
+
+const assertTXEq = (a: TransferOffchainTx, b: TransferOffchainTx) => {
```
ChihChengLiang, line 65: We can also compare the tx hashes
`assert.equal(a.hash(), b.hash())`

ts/client/txPool.ts
```diff
@@ -1,40 +1,159 @@
...
+        ];
+        return {
+            tokenID,
+            feeReceiverID,
+            sumFees: highValue
+        };
+    }
+
+    private getQueueValue(tokenIDStr: string): BigNumber {
+        return sum(this.tokenIDStrToQueue[tokenIDStr].map(tx => tx.fee));
```
ChihChengLiang, line 157: Note that we have a queue size of 1024.
We have a commitment size of 32.
We pack 1 commitment at a time. 
So if we have a token A queue of 32 txs each worth $10 and a token B queue of 1024 tx each worth $1. The `getQueueValue ` considers token B queue is more valuable but token A queue is more profitable in a single pack. 

Not a big problem to fix now though.

ChihChengLiang (commented):
Looks good to me

ts/client/config.ts
jacque006: We can do that, I see the random part as more test related but no harm in having that available in prod code.

ts/client/txPool.ts
jacque006: Ah good catch, that is something I can easily fix.

jacque006: Move to new issue

ts/client/config.ts
jacque006: https://github.com/thehubbleproject/hubble-contracts/issues/628

ts/client/txPool.ts
jacque006: https://github.com/thehubbleproject/hubble-contracts/issues/628

PR #622, opened 2021-07-12
Add error logging handler for provider

jacque006:
This will log out provider errors that were silent before.


jacque006:
FYI @ChihChengLiang @kautukkundan going to merge early for use in devnet

PR #630, opened 2021-07-19
Integrated db engine with states

kautukkundan:


test/client/integration.test.ts
```diff
@@ -160,5 +160,5 @@ describe("Client Integration", function() {
 
         assert.equal(storageSyncer.transactions.count(), maxTransfers);
         assert.equal(storagePacker.transactions.count(), maxTransfers);
-    }).timeout(300000);
+    }).timeout(600000);
```
kautukkundan, line 163: It is taking almost double the amount of time for running the test. Can we optimize this?

ts/client/features/transfer.ts
```diff
@@ -302,7 +302,7 @@ async function validateTransferStateTransition(
     validateSender(sender, tokenID, tx.amount, tx.fee);
     validateReceiver(receiver, tokenID);
     const senderKey = await storage.pubkey.get(sender.pubkeyID.toNumber());
-    if (tx.nonce != sender.nonce)
+    if (!tx.nonce.eq(sender.nonce))
```
jacque006, line 305: This shouldn't be showing up, likely need to rebase.

test/client/integration.test.ts
jacque006: Yes, we likely can bring this down by using [batch](https://github.com/Level/level#dbbatch-chained-form) operations in place of the one offs, especially when syncing or packing batches. We may be able to use the existing `StorageEngine` interface's `commit` function when writing data to do this easily with existing constructs.

We could also follow a similar pattern to what `geth` is doing and only write to disk every `n` seconds/minutes everything in storage. This opens up some data loss windows in crash/stop scenarios but could be worth performance benefits. We could also have proposer/packer workflows always write, but have syncing workflows only do the intermittent writes.

For now, I would make a new issue documenting this performance issue, and during one of our sync ups we can determine priority and possible break it down into smaller issues.

jacque006 (approved):
Asides from needed rebase with `master`, LGTM :+1:

Issue #629, opened 2021-07-19
Bug: Error: Bad nonce  tx 0  state 0


jacque006:
### Problem

```
     Error: Bad nonce  tx 0  state 0
      at validateTransferStateTransition (ts/client/features/transfer.ts:307:15)
      at pack (ts/client/features/transfer.ts:381:13)
      at packBatch (ts/client/features/transfer.ts:646:52)
      at TransferPackingCommand.packAndSubmit (ts/client/features/transfer.ts:691:51)
      at Packer.maybePackTransfers (ts/client/services/packer.ts:73:24)
      at Packer.runOnce (ts/client/services/packer.ts:90:9)
      at Context.<anonymous> (test/client/integration.test.ts:143:9)
```

### Triage

Regression of https://github.com/thehubbleproject/hubble-contracts/pull/621

https://github.com/thehubbleproject/hubble-contracts/blob/master/ts/client/features/transfer.ts#L305

Cannot use equality comparison (`==`, `!=`, ===`, `!==`) with BigNumbers as they are objects.

### Solution

Use `eq` with `BigNumber` for equality checks.

PR #631, opened 2021-07-19
Fix nonce equality check to use .eq

jacque006:
Resolves https://github.com/thehubbleproject/hubble-contracts/issues/629


jacque006:
FYI @ChihChengLiang @kautukkundan will merge after CI passes.

PR #635, opened 2021-07-22
Delay rpc calls until node synced

### LINKED ISSUE #558: starting up the rpc client until after the initial sync is completed.
### What's wrong

The RPC client shouldn't be serving before the node is fully synced.

### How can we fix it

Should return HTTP 503

TODO: Need to figure out how to implement this

kautukkundan:
closes #558

### DIFF
--- a/test/client/integration.test.ts
+++ b/test/client/integration.test.ts
@@ -160,5 +160,5 @@ describe("Client Integration", function() {
 
         assert.equal(storageSyncer.transactions.count(), maxTransfers);
         assert.equal(storagePacker.transactions.count(), maxTransfers);
-    }).timeout(600000);
+    }).timeout(900000);
 });
--- a/ts/client/node.ts
+++ b/ts/client/node.ts
@@ -5,7 +5,7 @@ import { FastifyInstance } from "fastify";
 import { storageManagerFactory } from "../factory";
 import * as mcl from "../../ts/mcl";
 import { Bidder } from "./services/bidder";
-import { SyncerService } from "./services/syncer";
+import { SyncerService, SyncMode } from "./services/syncer";
 import { Packer } from "./services/packer";
 import { TransferPool } from "./features/transfer";
 import { BurnAuctionWrapper } from "../burnAuction";
@@ -89,9 +89,16 @@ export class HubbleNode {
             throw new Error("watcher is currently not supported");
         }
 
-        // In the future, we will want to delay starting up the rpc client
-        // until after the initial sync is completed (HTTP 503).
-        // https://github.com/thehubbleproject/hubble-contracts/issues/558
+        fast.addHook("onRequest", async (_request, reply) => {
+            if (syncer.getMode() === SyncMode.INITIAL_SYNCING) {
+                return reply.status(503).send({
+                    message: "Initial sync incomplete",
+                    error: "RPC unavailable",
+                    statusCode: 503
+                });
+            }
+        });
+
         const rpc = new RPC(api, fast, transferPool);
         return new this(
             modes,


test/client/integration.test.ts
```diff
@@ -160,5 +160,5 @@ describe("Client Integration", function() {
 
         assert.equal(storageSyncer.transactions.count(), maxTransfers);
         assert.equal(storagePacker.transactions.count(), maxTransfers);
-    }).timeout(600000);
+    }).timeout(900000);
```
kautukkundan, line 163: increased time limit as tests keep failing at the edge of 10min

jacque006, line 163: Why was this needed?

ts/client/index.ts
```diff
@@ -15,6 +16,16 @@ async function main() {
...
 
+    fast.addHook("onRequest", async (request, reply) => {
+        if (node.getSyncMode() === SyncMode.INITIAL_SYNCING) {
+            return reply.status(503).send({
+                message: "Initial sync incomplete",
+                error: "RPC unavailable",
+                statusCode: 503
+            });
+        }
+    });
```
jacque006, line 27: I think we can move this into `node.ts` and remove the getter for syncMode, more appropriate for node to manage this hook.

I addition, one downside of this is that all endpoints will return this status including our process health check (`/health`) which would be used by container orchestration platforms and uptime checks in general to make sure the node is still responding to pings. We should have the hook ignore `/health` or consider adding the hook only to the endpoints registered in `RPC.ts`.

kautukkundan: the fastify Instance is only present inside the "static" init method of the node class, adding the onRequest hook inside that function won't work at runtime as we need to check the syncMode at every API call. to make node manage the hook, fastify instance must be saved as a class variable and then the hook must be added inside the `start` method.

jacque006: You can move to node without saving instance since the `syncer` service is in scope:
Screenshot from 2021-07-22 17-41-53

kautukkundan: oh so sorry, my bad
i was thinking that the `syncer` instance gets destroyed outside the scope of this init function. thanks 😃

kautukkundan:
Some notes -

1. Based on Jake's previous comment
> I addition, one downside of this is that all endpoints will return this status including our process health check (/health) which would be used by container orchestration platforms and uptime checks in general to make sure the node is still responding to pings. We should have the hook ignore /health or consider adding the hook only to the endpoints registered in RPC.ts.

I guess this is sorted as the fastify hook only applies to the fastify endpoints. since the `/health` endpoint is created with the terminus, it does not have the `onRequest` hook applied to it. I have manually tested this as well. 

2. I tried to write a test case for testing out the rpc endpoints. However, the server requires geth node to be running already on port 8545 with the contracts deployed and registered. This needs some discussion if automated testing is required for the api right now. Also should that be included in the existing test suite or a separate test suite specifically for testing out APIs which runs separately after the default test suite.

ts/client/node.ts
```diff
@@ -89,9 +89,18 @@ export class HubbleNode {
             throw new Error("watcher is currently not supported");
         }
 
-        // In the future, we will want to delay starting up the rpc client
-        // until after the initial sync is completed (HTTP 503).
-        // https://github.com/thehubbleproject/hubble-contracts/issues/558
+        fast.addHook("onRequest", async (_request, reply) => {
+            console.log("hook ran");
+
```
jacque006, line 94: Remove

jacque006 (approved):
Asides from errant console.log, looks good

ts/client/node.ts
kautukkundan: oops! missed this

jacque006:
For testing, given the amount or work it will involve to setup I think we can skip for now. Manual testing is good :+1:

Issue #637, opened 2021-07-29
Include subtreeID and depositCount in DepositQueued event in IDepositManager


jacque006:
### Problem

When a deposit is queued using `DepositManager.depositFor`, it is difficult as a contract consumer to determine what subtree the deposit will become part of and how many deposits are currently part of that subtree. This also makes it difficult to correlate the deposit to the `DepositsFinalised` event from `Rollup.submitDeposits`.

### Solution

Add `uint256 subtreeID` and `uint256 depositCount` to the `IDepositManager.DepositQueued` event.

PR #638, opened 2021-07-29
Add subtreeID, depositCount to DepositQueued event

jacque006:
Resolves https://github.com/thehubbleproject/hubble-contracts/issues/637

### DIFF
--- a/contracts/DepositManager.sol
+++ b/contracts/DepositManager.sol
@@ -11,7 +11,13 @@ import { Rollup } from "./rollup/Rollup.sol";
 import { ITokenRegistry } from "./TokenRegistry.sol";
 
 interface IDepositManager {
-    event DepositQueued(uint256 pubkeyID, uint256 tokenID, uint256 l2Amount);
+    event DepositQueued(
+        uint256 pubkeyID,
+        uint256 tokenID,
+        uint256 l2Amount,
+        uint256 subtreeID,
+        uint256 depositID
+    );
     event DepositSubTreeReady(uint256 subtreeID, bytes32 subtreeRoot);
 
     function paramMaxSubtreeDepth() external returns (uint256);
@@ -23,6 +29,9 @@ interface IDepositManager {
     function reenqueue(bytes32 subtreeRoot) external;
 }
 
+/**
+ * @dev subtreeID starts at 1
+ */
 contract SubtreeQueue {
     // Each element of the queue is a root of a subtree of deposits.
     mapping(uint256 => bytes32) public queue;
@@ -31,8 +40,8 @@ contract SubtreeQueue {
 
     event DepositSubTreeReady(uint256 subtreeID, bytes32 subtreeRoot);
 
-    function enqueue(bytes32 subtreeRoot) internal {
-        uint256 subtreeID = back + 1;
+    function enqueue(bytes32 subtreeRoot) internal returns (uint256 subtreeID) {
+        subtreeID = back + 1;
         back = subtreeID;
         queue[subtreeID] = subtreeRoot;
         emit DepositSubTreeReady(subtreeID, subtreeRoot);
@@ -65,9 +74,12 @@ contract DepositCore is SubtreeQueue {
         paramMaxSubtreeSize = 1 << maxSubtreeDepth;
     }
 
-    function insertAndMerge(bytes32 depositLeaf) internal {
-        depositCount++;
-        uint256 i = depositCount;
+    function insertAndMerge(bytes32 depositLeaf)
+        internal
+        returns (uint256 subtreeID, uint256 depositID)
+    {
+        depositID = depositCount + 1;
+        uint256 i = depositID;
 
         uint256 len = babyTreesLength;
         babyTrees[len] = depositLeaf;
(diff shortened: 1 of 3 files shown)


contracts/DepositManager.sol
```diff
@@ -83,9 +95,13 @@ contract DepositCore is SubtreeQueue {
             i >>= 1;
         }
         babyTreesLength = len;
+        // TODO Is there a better way we can mirror this as it is in enqueue function?
+        subtreeID = back + 1;
```
jacque006, line 99: Is there a better way we can do this to reduce duplication of next subtreeID being calculated?

ts/client/services/events/depositQueued.ts
```diff
@@ -13,7 +13,7 @@ export class DepositQueuedEventSyncer extends ContractEventSyncer {
     constructor(api: CoreAPI) {
         super(
             api.contracts.depositManager,
-            api.contracts.depositManager.filters.DepositQueued(null, null, null)
+            api.contracts.depositManager.filters.DepositQueued()
```
jacque006, line 16: Newer typechain/ethers versions do not require `null` filling. Will update this where appropriate in future work, and may do a tech-debt PR after public testnet to go through client/tests and update all of these.

jacque006:
Gas changes

### master
Screenshot from 2021-07-29 11-39-53
Deposit 0: `139601`
Deposit 1: `109921`

### This PR
Screenshot from 2021-07-29 11-40-40
Deposit 0: `142333`
Deposit 1: `110665`
Deposit 2: `108145`

### cha-cha-cha-changes

Deposit 0 change: `+2732`
Deposit 0 % change: `1.957%`

Deposit 1 change: `+744`
Deposit 1 % change: `0.677%`

Deposit 2 N/A

contracts/DepositManager.sol
ChihChengLiang: Yes, let me reply in other comments

contracts/DepositManager.sol
```diff
@@ -65,7 +74,10 @@ contract DepositCore is SubtreeQueue {
         paramMaxSubtreeSize = 1 << maxSubtreeDepth;
     }
 
-    function insertAndMerge(bytes32 depositLeaf) internal {
+    function insertAndMerge(bytes32 depositLeaf)
+        internal
+        returns (uint256 subtreeID, uint256 prevDepositCount)
+    {
         depositCount++;
         uint256 i = depositCount;
```
ChihChengLiang, line 82: We make good use of the loaded depositCount here.
```suggestion
        depositID = depositCount + 1; // 1 SLOAD here
        uint256 i = depositID;
```

contracts/DepositManager.sol
```diff
@@ -83,9 +95,13 @@ contract DepositCore is SubtreeQueue {
             i >>= 1;
         }
         babyTreesLength = len;
+        // TODO Is there a better way we can mirror this as it is in enqueue function?
+        subtreeID = back + 1;
+        prevDepositCount = depositCount;
```
ChihChengLiang, line 100: Don't do them here yet.
```suggestion
```

contracts/DepositManager.sol
```diff
@@ -65,7 +74,10 @@ contract DepositCore is SubtreeQueue {
         paramMaxSubtreeSize = 1 << maxSubtreeDepth;
     }
 
-    function insertAndMerge(bytes32 depositLeaf) internal {
+    function insertAndMerge(bytes32 depositLeaf)
+        internal
+        returns (uint256 subtreeID, uint256 prevDepositCount)
```
ChihChengLiang, line 79: Rename to make sense the following feedbacks.
```suggestion
        returns (uint256 subtreeID, uint256 depositID)
```

contracts/DepositManager.sol
```diff
@@ -83,9 +95,13 @@ contract DepositCore is SubtreeQueue {
...
+        subtreeID = back + 1;
+        prevDepositCount = depositCount;
+
         // Subtree is ready, send to SubtreeQueue
         if (depositCount == paramMaxSubtreeSize) {
-            enqueue(babyTrees[0]);
+            subtreeID = enqueue(babyTrees[0]);
             // reset
             babyTreesLength = 0;
             depositCount = 0;
```
ChihChengLiang, line 107: Do early return here.
```suggestion
            depositCount = 0;
            return
```
Then after the "if" block, we do 
```
// The case of no reset
subtreeID = back + 1;
// bump the depositCount
depositCount = depositID;
```

contracts/DepositManager.sol
```diff
@@ -83,9 +95,13 @@ contract DepositCore is SubtreeQueue {
             i >>= 1;
         }
         babyTreesLength = len;
+        // TODO Is there a better way we can mirror this as it is in enqueue function?
+        subtreeID = back + 1;
+        prevDepositCount = depositCount;
+
         // Subtree is ready, send to SubtreeQueue
         if (depositCount == paramMaxSubtreeSize) {
```
ChihChengLiang, line 103:
```suggestion
        if (depositID == paramMaxSubtreeSize) {
```

ChihChengLiang (commented):
Proposed a way to assign subtreeID gracefully.

jacque006:
Thanks for the help @ChihChengLiang ! Suggestions implemented, gas stats updated, slight decrease :tada:

ChihChengLiang (commented):
LGTM

kautukkundan:
Suggestion - we can start using [eth-gas-reporter](https://github.com/cgewecke/eth-gas-reporter/). It also comes with [codechecks](https://github.com/cgewecke/eth-gas-reporter/blob/master/docs/codechecks.md) plug-in which can compare gas change with the target branch so we don't need to calculate manually!

jacque006:
@kautukkundan I have a branch up that would add contract test coverage as well as hopefully implement the reporter you mentioned https://github.com/thehubbleproject/hubble-contracts/pull/552 . We would need to update a number of spots for it to work properly but I agree that it would be much easier than manually calculating and reporting in PRs :)

PR #639, opened 2021-07-31
updated deposit manager to skip erc20 transfer

### LINKED ISSUE #627: Skip ERC20 allowance check and transfer when depositFor l1Amount is zero
### Problem

When a user calls `DepositManager.depositFor` with a zero (`0`) l1Amount, a ERC20 check and then transfer are called (https://github.com/thehubbleproject/hubble-contracts/blob/master/contracts/DepositManager.sol#L155-L159). This is unnecessary and leads to a higher gas cost.

### Solution

 - Skip https://github.com/thehubbleproject/hubble-contracts/blob/master/contracts/DepositManager.sol#L155-L159 when `l1Amount` is `0`.
 - Add test case for the above to https://github.com/thehubbleproject/hubble-contracts/blob/master/test/fast/deposit.test.ts , as well as test cases for `require` failures in `depositFor`.

kautukkundan:
- [x] updated deposit manager contract to skip erc20 check when amount zero
- [x] added test cases to test above functionality + other revert statements
 
closes #627

### DIFF
--- a/contracts/DepositManager.sol
+++ b/contracts/DepositManager.sol
@@ -169,12 +169,15 @@ contract DepositManager is
             "l1Amount should be a multiple of l2Unit"
         );
         // transfer from msg.sender to vault
-        require(
-            IERC20(addr).allowance(msg.sender, address(this)) >= l1Amount,
-            "token allowance not approved"
-        );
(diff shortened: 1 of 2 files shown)


ChihChengLiang (commented):
LGTM

jacque006 (approved):

PR #640, opened 2021-07-31
added RollbackTriggered event

### LINKED ISSUE #381: Emit an event in startRollingBack
### What's wrong

When people dispute a batch. We could only learn the event from the calldata. It might take a while to learn what and why a dispute happened.

### How can we fix it

We can add an event `RollbackTriggered(batchID, result)` to better know what's happening when a dispute happened.

kautukkundan:
closes #381

### DIFF
--- a/contracts/rollup/BatchManager.sol
+++ b/contracts/rollup/BatchManager.sol
@@ -39,6 +39,7 @@ contract BatchManager {
     event NewBatch(uint256 batchID, bytes32 accountRoot, Types.Usage batchType);
     event StakeWithdraw(address committed, uint256 batchID);
     event RollbackStatus(uint256 startID, uint256 nDeleted, bool completed);
+    event RollbackTriggered(uint256 batchID, Types.Result result);
 
     modifier isNotRollingBack() {
         require(invalidBatchMarker == 0, "BatchManager: Is rolling back");
@@ -88,8 +89,11 @@ contract BatchManager {
         batch = batches[batchID];
     }
 
-    function startRollingBack(uint256 invalidBatchID) internal {
+    function startRollingBack(uint256 invalidBatchID, Types.Result result)
+        internal
+    {
         invalidBatchMarker = invalidBatchID;
+        emit RollbackTriggered(invalidBatchID, result);
         rollback();
     }
 
--- a/contracts/rollup/Rollup.sol
+++ b/contracts/rollup/Rollup.sol
@@ -390,7 +390,7 @@ contract Rollup is BatchManager, EIP712, IEIP712 {
         if (
             result != Types.Result.Ok ||
             (processedStateRoot != target.commitment.stateRoot)
-        ) startRollingBack(batchID);
+        ) startRollingBack(batchID, result);
     }
 
     function disputeTransitionMassMigration(
@@ -419,7 +419,7 @@ contract Rollup is BatchManager, EIP712, IEIP712 {
         if (
             result != Types.Result.Ok ||
             (processedStateRoot != target.commitment.stateRoot)
-        ) startRollingBack(batchID);
+        ) startRollingBack(batchID, result);
     }
 
     function disputeTransitionCreate2Transfer(
@@ -449,7 +449,7 @@ contract Rollup is BatchManager, EIP712, IEIP712 {
         if (
             result != Types.Result.Ok ||
             (processedStateRoot != target.commitment.stateRoot)
-        ) startRollingBack(batchID);
+        ) startRollingBack(batchID, result);
     }
 
     function disputeSignatureTransfer(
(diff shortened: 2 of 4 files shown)


contracts/rollup/Rollup.sol
```diff
@@ -419,7 +422,10 @@ contract Rollup is BatchManager, EIP712, IEIP712 {
         if (
             result != Types.Result.Ok ||
             (processedStateRoot != target.commitment.stateRoot)
-        ) startRollingBack(batchID);
+        ) {
+            emit RollbackTriggered(batchID, result);
+            startRollingBack(batchID);
+        }
```
kautukkundan, line 428: right now i have added this `emit` statement wherever the `startRollingBack` function is called. Another way could be to emit this event inside the `startRollingBack` function itself, but that would require a change in the function signature to include the `result` as an argument and subsequent changes wherever this function is called. 
Is the current approach fine?

ChihChengLiang: Good question. I would vote to change the function signature of startRollingBack since the RollbackTriggered should be emitted wherever startRollingBack is called. It also makes sense for the startRollingBack to take the result as an argument, as `result` tells the `reason` why we start rolling back.

contracts/test/TestRollup.sol
```diff
@@ -52,7 +52,7 @@ contract TestRollup is BatchManager {
 
     function testRollback(uint256 batchID) external returns (uint256) {
         uint256 g = gasleft();
-        startRollingBack(batchID);
+        startRollingBack(batchID, Types.Result.BadSignature);
```
kautukkundan, line 55: provided a one off default "bad" result as this is a mock contract

test/fast/rollback.test.ts
```diff
@@ -52,7 +52,11 @@ describe("Rollback", function() {
             rollup.filters.RollbackStatus(null, null, null),
             tx.blockHash
         );
-
+        const [event] = await rollup.queryFilter(
+            rollup.filters.RollbackTriggered(null, null),
```
jacque006, line 56: You should be able to remove these extraneous nulls here and in other filters, see https://github.com/thehubbleproject/hubble-contracts/pull/638#discussion_r679356273

test/fast/rollback.test.ts
```diff
@@ -65,24 +69,36 @@ describe("Rollback", function() {
...
+        const batchID = 1;
+        const tx = await rollup.testRollback(batchID);
         const [status] = await rollup.queryFilter(
             rollup.filters.RollbackStatus(null, null, null),
             tx.blockHash
         );
+        const [event] = await rollup.queryFilter(
+            rollup.filters.RollbackTriggered(null, null),
+            tx.blockHash
+        );
```
jacque006, line 81: When multiple async (promise) calls are made that don't have data dependencies on each other, it can be a good idea to let them run in parallel especially when the underlying async function is making RPC calls. Note that this should not be done for L1 txns/writes. This can also be done in a few other spots in this test, and can reduce test runtimes.

```suggestion
        const [[status], [event]] = await Promise.all([
            rollup.queryFilter(
                rollup.filters.RollbackStatus(),
                tx.blockHash
            ),
            rollup.queryFilter(
            	rollup.filters.RollbackTriggered(),
            	tx.blockHash
            )
        ]);
```

jacque006 (approved):
Couple of suggestions, but ok to go out as-is :+1:

test/fast/rollback.test.ts
kautukkundan: good advice, both of them, will implement 👍🏼

jacque006 (approved):

contracts/rollup/BatchManager.sol
```diff
@@ -39,6 +39,7 @@ contract BatchManager {
     event NewBatch(uint256 batchID, bytes32 accountRoot, Types.Usage batchType);
     event StakeWithdraw(address committed, uint256 batchID);
     event RollbackStatus(uint256 startID, uint256 nDeleted, bool completed);
+    event RollbackTriggered(uint256 batchID, Types.Result result);
```
ChihChengLiang, line 42: This is a comment, no action is required.

We have two events that will be emitted in the startRollingBack, let me explain why we need two. The short answer is RollbackTriggered is only called once per dispute and the RollbackStatus is called once per rollback.

For example, we have a long rollback that takes too much gas so we need to slash in multiple steps. Say the tip batch is 120, the disputed batch is 100, and then the last good batch is 99.

- The first step we submit the fraud-proof and determine if we need to do startRollingBack. The fraud-proof part might already takes a lot of gas. The startRollingBack is initiated and then it emits RollbackTriggered, which tells us 100 is disputed and determined invalid for BadSignature reason. startRollingBack then calls rollback
- this rollback() only rolls back from 120 to 110 (110 not included), which emits RollbackStatus telling us it "started" from batchID 120, "deleted" n = 10 batches, and not "completed" yet.
- some other people called keepRollingBack. It now rolls back from 110 to 100. RollbackStatus tells us this time it "started" from batchID 110, "deleted" n = 10 batches, and "completed" the slash.

ChihChengLiang (commented):
LGTM.

PR #644, opened 2021-08-05
Fix docker release pipeline

jacque006:


jacque006:
FYI @kautukkundan @ChihChengLiang

Issue #651, opened 2021-08-10
Bug: MultiTokenPool.getHighestValueToken fails if queue only has transfers with fee of 0


jacque006:
### Problem

When a user/account sends a transfer tx with a fee of 0, `MultiTokenPool` is unable to properly prioritize and retrieve that transfer if a client node is feeling charitable and will pack that transfer.

Note that this unlikely/an edge case, but does appear to be supported by the node and the Hubble protocol.

### Solution

Update `MultiTokenPool.getHighestValueToken` to support queues with a summed fee value of zero. Also add a specific error case if that function is unable to determine the most valuable queue to process.

jacque006:
@ChihChengLiang Your input would be appreciated if a tx having a fee of 0 is supported by the protocol.

PR #652, opened 2021-08-10
Fix getHighestValueToken to work with zero fee txs

jacque006:
Resolves https://github.com/thehubbleproject/hubble-contracts/issues/651


kautukkundan (approved):
All good 👍🏼

PR #657, opened 2021-08-30
implemented event listener for `BatchPubkeyRegistered` event

### LINKED ISSUE #625: Implement BatchPubkeyRegisteredEventSyncer
### Problem

Syncing of batch pubkey registrations to the client is currently not implemented. See `BatchPubkeyRegisteredEventSyncer`.

### Solution

Implement batch pubkey registration syncing in client. See `SinglePubkeyRegisteredEventSyncer` as a reference.

kautukkundan:
closes #625

### DIFF
--- a/test/client/integration.test.ts
+++ b/test/client/integration.test.ts
@@ -80,6 +80,19 @@ describe("Client Integration", function() {
         const domainSeparator = await contracts.rollup.domainSeparator();
         group.setupSigners(arrayify(domainSeparator));
 
+        const pubkeyBatch: any = [];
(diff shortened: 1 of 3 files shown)


jacque006 (changes requested):
We should also add batch pubkey registration into the client integration test to confirm this is working, similar to registering a single pubkey https://github.com/thehubbleproject/hubble-contracts/blob/master/test/client/integration.test.ts#L93 . We can have half the keys be single, the other half as batch.

jacque006 (approved):

Issue #666, opened 2021-09-23
parameterize construction of database connection


kautukkundan:
### Problem

Currently database objects are simply created at a fixed location and passed on to all the consumers directly. This also creates overlapping of read operations especially in the case when multiple storage managers access the database in the same context (see `client/integration.test.ts`) 

### Solution

create a factory pattern for database connections which can be modified according to context

Issue #624, opened 2021-07-13
make path of levelDB database configurable


kautukkundan:
currently the path of levelDB folder is hardcoded. 
additionally - it'll better to have separate path for DB while running tests so that it doesn't interfere with DB running in production. 

https://github.com/thehubbleproject/hubble-contracts/pull/619#discussion_r668291050_

PR #667, opened 2021-09-23
Persist transactions in database

### LINKED ISSUE #660: Persist TransactionStorage
Previously: https://github.com/thehubbleproject/hubble-contracts/issues/548

### Problem

`TransactionStorage` is currently implemented in memory and does not maintain state between process crashes/restarts.

### Solution

Implement `TransactionStorage` using leveldb to persist L2 transaction info and metadata.

kautukkundan:
closes #660 

Did not update storageManagerFactory yet. Will update that and integration tests when addressing #624 and #666. 
wdyt?

### DIFF
--- a/test/client/transaction.test.ts
+++ b/test/client/transaction.test.ts
@@ -1,16 +1,20 @@
 import { BigNumber } from "@ethersproject/bignumber";
+import { arrayify } from "@ethersproject/bytes";
 import chai, { assert } from "chai";
 import chaiAsPromised from "chai-as-promised";
+import del from "del";
+import { BlsSigner } from "../../ts/blsSigner";
 import { OffchainTx } from "../../ts/client/features/interface";
 import { TransferOffchainTx } from "../../ts/client/features/transfer";
 import { Status } from "../../ts/client/storageEngine/transactions/constants";
-import { TransactionStorage } from "../../ts/client/storageEngine/transactions/interfaces";
-import { TransactionMemoryStorage } from "../../ts/client/storageEngine/transactions/memory";
+import { TransactionDBStorage } from "../../ts/client/storageEngine/transactions/db";
 import {
     StatusTransitionInvalid,
     TransactionAlreadyExists,
     TransactionDoesNotExist
 } from "../../ts/exceptions";
+import * as mcl from "../../ts/mcl";
+import { randHex } from "../../ts/utils";
 
 chai.use(chaiAsPromised);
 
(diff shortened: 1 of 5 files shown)


test/client/transaction.test.ts
```diff
@@ -99,21 +110,21 @@ describe("TransactionMemoryStorage", () => {
             });
 
             it("properly transitions to failed state from pending", async function() {
-                const txn = txFactory(10, 11, 10101, 101, 0);
+                const txn = txFactory(10, 11, 1010, 101, 0);
```
kautukkundan, line 113: https://github.com/thehubbleproject/hubble-contracts/blob/tx2db/ts/decimal.ts#L69 is causing the issue so reduced the value. Unsure why it used to pass before.

ts/client/storageEngine/transactions/db.ts
```diff
@@ -0,0 +1,197 @@
...
+        return txnStatus;
+    }
+
+    public async count(): Promise<number> {
+        const stream = this.db.createKeyStream();
+        let count = 0;
+        for await (const _ of stream) {
+            count++;
+        }
+        return count;
```
jacque006, line 104: Another way we could do this is to store the count at a static key. This would require us to update the count on every insert (increased write time), but would make reads very quick. And in the worst case on a key miss we would have to do this count anyways. Would look like:

1. Check for count key. If exists return value.
2. If value missing at key, run count from stream. Write result and return.

In addition to needing to increment count when inserting new values. Think we can skip this optimization for now and consider it in the future if this function becomes a bottleneck.

ts/client/storageEngine/transactions/db.ts
```diff
@@ -0,0 +1,197 @@
...
+
+    private async getStatusOrFail(message: string): Promise<TransactionStatus> {
+        try {
+            return await this.get(message);
+        } catch (error) {
+            if (error.name === "NotFoundError") {
+                throw new TransactionDoesNotExist(message);
+            } else {
+                throw error;
+            }
```
jacque006, line 135: #nitpick `else` block is not needed here. Same as:
```typescript
            if (error.name === "NotFoundError") {
                throw new TransactionDoesNotExist(message);
            }
            throw error;
```
I generally prefer the above form to reduce nesting but it is a stylistic choice. Good as is.

ts/client/storageEngine/transactions/db.ts
```diff
@@ -0,0 +1,197 @@
...
+    private async fromDB(msg: string): Promise<TransactionStatus> {
+        const object = JSON.parse(await this.db.get(msg));
+
+        return {
+            transaction: TransferOffchainTx.deserialize(arrayify(object.tx)),
+            status: object.status,
+            detail: object.detail || undefined,
+            batchID: object.batchID || undefined,
+            l1TxnHash: object.l1TxnHash || undefined,
+            l1BlockIncluded: object.l1BlockIncluded || undefined
```
jacque006, line 194: Are the `|| undefined`'s needed? Is there something else that needs for the field not to be `null` or some other falsy value?

jacque006 (approved):
Nice work, good idea to split DB factory into another issue/PR. Few observational comments, :shipit:

ts/client/storageEngine/transactions/db.ts
kautukkundan: Ah right, these values will be undefined implicitly anyway. No need of `|| undefined`

ts/client/storageEngine/transactions/db.ts
kautukkundan: I did explore this when I was trying to save Batches in the DB (upcoming PR). There is no inherent issue with this (infact it has some advantages like you mentioned) but saving `count` in a db structure with `key = msg` and `value = serialized tx` felt a little unintuitive to me as the db was not "meant to" store that information. Also something which is enforced in the SQL land, keeping a db object pure.

Open to interpretation though.

jacque006: We could always add a new 'table' (subdb) that is just for storing counts with the key being related to the entity it is storing if we wanted to have that explicit separation. I don't think there is a ton of harm in a k/v store of having different values being stored in the same area, as the key delimiters can be used to separate types of data.

For example, if we had a Cat area (subdb) and we also wanted to include a counter cache for it, we might key it like:
```
cat_123: ... // cat 1 serialized data
cat_456: ... // cat 2 serialized data
cat_num_cats: 2
``` 
And then having a TS const like `NUM_CATS_KEY = "num_cats"`. It would be very clear in code when writes/reads are modifying the count and not a kitty cat.

All this being said, don't think we need this level of per-optimzation yet and we could come back to it. Feel free to resolve.

kautukkundan: Right makes sense. Thanks 😄 
Just need to make note of not including the cache/metadata per db when doing things like counting keys etc as `createStream` will include these keys as well.

Issue #671, opened 2021-11-10
Inconsistent deposit subtree IDs after rollback


b-tarczynski:
Right now rollbacks of deposit batches call `reenqueue` function on `depositManager` that creates new subtree with different subtreeID than was used in `submitDeposits` and emits new `DepositSubTreeReady` event with this ID.
That is a problem in the case of syncing Deposits and DepositSubtrees from events. 
If we sync in commander `DepositQueued` events and add them to queue and we pop Deposits on `DepositSubTreeReady` events in case of rollback we wouldn't know which `DepositQueued` events assign to `DepositSubTreeReady` that was emitted for `reenqueue` function.

Way to fix that would be to implement the `pushFront(bytes32 subtreeRoot)` function that adds rollbacked Subtree at the beginning of the queue with the same SubtreeID.
On the command side we would simply check if we already store subtree with this ID and in that case we would omit it.
If we want to go with that approach then we should rename other `SubtreeQueue` functions, for example:
```
enqueue -> pushBack(bytes32 subtreeRoot)
dequeue -> popFront() returns (uint256 subtreeID, bytes32 subtreeRoot)
```

Also it would make deposits stay in original order after rollback, because right now submitted deposits are added at the end of the queue.

jacque006:
Overall this solution sounds good to me. The only confusing part is that there may now be multiple `DepositSubTreeReady` with the same args, but that can be pointed out/documented in `SubtreeQueue`.

> If we want to go with that approach then we should rename other SubtreeQueue functions, for example:
enqueue -> pushBack(bytes32 subtreeRoot)
dequeue -> popFront() returns (uint256 subtreeID, bytes32 subtreeRoot)

I could see the new function just being called `requeue` to match original authors naming. If we wanted to follow ECMA/JS syntax which Solidity largely appears to, we would have `push`,  `shift`, and `unshift`. But frankly, they all get the intent across :)

@ChihChengLiang feel free to chime in if you have any additional thoughts on this.

ChihChengLiang:
The solution sounds good to me too. I like the `pushBack`, `pushFront`, and `popFront` namings too.

PR #672, opened 2021-11-18
Inconsistent deposit subtree IDs after rollback

b-tarczynski:
Resolves https://github.com/thehubbleproject/hubble-contracts/issues/671

Decided to follow naming proposed by @jacque006 and not to emit duplicated `DepositSubTreeReady` event on  `reenqueue` function.
Let me know if you find any places for improvement.

### DIFF
--- a/test/slow/rollup.deposit.test.ts
+++ b/test/slow/rollup.deposit.test.ts
@@ -1,0 +1,153 @@
+import { deployAll } from "../../ts/deploy";
+import { TESTING_PARAMS } from "../../ts/constants";
+import { ethers } from "hardhat";
+import { StateTree } from "../../ts/stateTree";
+import { AccountRegistry } from "../../ts/accountTree";
+import { serialize } from "../../ts/tx";
+import * as mcl from "../../ts/mcl";
+import { allContracts } from "../../ts/allContractsInterfaces";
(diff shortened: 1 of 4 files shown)


test/slow/rollup.deposit.test.ts
```diff
@@ -0,0 +1,195 @@
...
+
+        const { exampleToken, depositManager } = contracts;
+        erc20 = new ERC20ValueFactory(await exampleToken.decimals());
+        await exampleToken.approve(
+            depositManager.address,
+            erc20.fromHumanValue("1000000").l1Value
+        );
+    });
+
+    it.only("reenqueue deposit subtree on rollback", async function() {
```
jacque006, line 81:
```suggestion
    it("reenqueue deposit subtree on rollback", async function() {
```

jacque006 (changes requested):
Contract changes look good to me. @ChihChengLiang if you have time your review of the `DepositManager.sol` changes would be appreciated.

Although I see some benefits of the thoroughness of the new slow deposit test, it will increase the overall time the full test suite will take to run. Was there a reason you opted for the new slow test instead of modifying the existing `Test rollback with deposits` test case?

jacque006 (approved):

PR #674, opened 2021-12-03
Make mass migration tx type consistent with batch type

duckception:
`Transfer` and `Create2Transfer` transaction types are consistent with their corresponding bach types (`1`  for `Transfer` and `3` for `Create2Transfer`) while `MassMigration` types are inconsistent i.e., the transaction type is set to `5` while the batch type is set to `2`. 

This PR modifies the `MassMigration` transaction type so it corresponds to its batch type.

### DIFF
--- a/contracts/libs/Tx.sol
+++ b/contracts/libs/Tx.sol
@@ -14,8 +14,8 @@ import { BLS } from "./BLS.sol";
 library Tx {
     // Tx types in uint256
     uint256 private constant TRANSFER = 1;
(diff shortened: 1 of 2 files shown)


jacque006:
This looks like a valid correction to me, but want to wait on confirmation from @ChihChengLiang

ChihChengLiang (commented):
Nice catch! 
The fix looks good to me.

jacque006 (approved):

PR #676, opened 2021-12-20
Add `SpokeRegistered` event

max-sidorov:
Added an event that emits after the registration of spokes.

### DIFF
--- a/contracts/SpokeRegistry.sol
+++ b/contracts/SpokeRegistry.sol
@@ -2,12 +2,15 @@
 pragma solidity ^0.6.12;
 
 contract SpokeRegistry {
+    event RegisteredSpoke(uint256 spokeID, address spokeContract);
+
     mapping(uint256 => address) public registeredSpokes;
     uint256 public numSpokes;
 
     function registerSpoke(address spokeContract) external {
         numSpokes++;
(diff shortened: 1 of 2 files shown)


msieczko (changes requested):
I'd rename `RegisteredSpoke` -> `SpokeRegistered`. Better grammar and more consistent with other events.

msieczko:
This PR would make syncing registered spokes much easier. @jacque006 could you take a look?

contracts/SpokeRegistry.sol
```diff
@@ -2,12 +2,15 @@
...
+    event SpokeRegistered(uint256 spokeID, address spokeContract);
+
     mapping(uint256 => address) public registeredSpokes;
     uint256 public numSpokes;
 
     function registerSpoke(address spokeContract) external {
         numSpokes++;
         registeredSpokes[numSpokes] = spokeContract;
+        emit SpokeRegistered(numSpokes, spokeContract);
     }
```
jacque006, line 14: Would be nice to have a fast test case for this function and the new event, but given that there is not an existing case will leave at author's discretion.

jacque006 (approved):

msieczko (approved):

jacque006 (approved):

Issue #584, opened 2021-05-04
Access control for contracts


ChihChengLiang:
### What's wrong

Some of these contracts have no access controls. Which might be a potential attack vector. This issue consolidate those contracts

#### SpokeRegistry

SpokeRegistry register spoke contracts and maps a spokeID to a spoke contract. Once a spoke is registered, people can send massMigration to migrate their tokens from Hubble to the spoke.

No access control means anyone can register malicious spoke contracts.

#### tokenRegistry

No access control means anyone can register malicious ERC20 token.

### How can we fix it

We don't have governance and we don't want to assign that access to a particular entity.

One way to fix this is to determine all the spokes and the tokens at the deployment and no one are allowed to add new tokens in the future.

PR #677, opened 2021-12-21
Modify TokenRegistry contract

max-sidorov:
Simplify registration to one tx instead of two and prevent registering the same token twice.

### DIFF
--- a/contracts/TokenRegistry.sol
+++ b/contracts/TokenRegistry.sol
@@ -4,16 +4,13 @@ import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 
 interface ITokenRegistry {
     event RegisteredToken(uint256 tokenID, address tokenContract);
-    event RegistrationRequest(address tokenContract);
 
     function safeGetRecord(uint256 tokenID)
-        external
-        view
-        returns (address, uint256 l2Unit);
+    external
+    view
+    returns (address, uint256 l2Unit);
 
-    function requestRegistration(address tokenContract) external;
-
-    function finaliseRegistration(address tokenContract) external;
+    function registerToken(address tokenContract) external;
 }
 
 contract TokenRegistry is ITokenRegistry {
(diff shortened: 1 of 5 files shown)


msieczko (changes requested):
I'd rename `RegisteredToken` -> `TokenRegistered` similarly as here https://github.com/thehubbleproject/hubble-contracts/pull/676#pullrequestreview-837549307. Other than that LGTM.

msieczko (approved):
LGTM. @jacque006 could you take a look as well and merge the PR?

jacque006:
Couple of things worth considering with these changes:
- I believe it was a past intent to in some way limit the registration of tokens to either deploy time or some sort of access control (see `tokenRegistry` in https://github.com/thehubbleproject/hubble-contracts/issues/584). For example, anyone might be able to request a token registration but only an owner/admin could finalize the registration.
- With these changes (and even before), anyone can register a token to a Hubble network. This could be nice as it allows anyone to onboard at token at their own purview/cost, but also means more scammy/less reputable tokens can also be registered which could have impacts on the reputation of the network. If having token registration fully open is a concern, I would at the very least recommend adding `@openzepplin/contracts`'s `Ownable` to the `TokenRegistry` and adding the `onlyOwner` modifier to the `registerToken`.

I have no strong opinion on either your changes or the access control schemes mentioned above as they all have their tradeoffs. Since you are the most active devs I will leave it at your discretion. Current changes look good :+1: assuming the above has been considered.

msieczko:
Hey @jacque006, thanks for your answer.
> Current changes look good 👍 assuming the above has been considered.
// yes I believe we need to protect tokenregistry and spokeregistry
// this PR does only simplifies the already open token registration 
// the other change is making sure that the same token is not registered twice which ma

msieczko:
Hey @jacque006, thanks for your answer 👍

> Current changes look good 👍 assuming the above has been considered.

I agree that we should to protect `TokenRegistry` (and `SpokeRegistry` similarly), so that possibly-malicious tokens are not added arbitrarily. Optimally I think it should be the responsibility of a DAO to accept/reject token registrations. 

We're not there yet though, so this PR is just meant to simplify the token registration to a single tx - it is open to anyone now and will still be. The second, more important change in this PR is preventing repeated registration of a token contract - currently `finaliseRegistration` can be called multiple times for the same token.

I suggest merging this PR now and tackling the issues that you mention in the future.

PR #678, opened 2021-12-28
Pass correct result on dispute invalid post state root 

b-tarczynski:
It would be nice to emit `RollbackTriggered` with correct result when rollback is triggered by invalid post state root(right now  it's `Types.Result.Ok`).

I added simple if statement in dispute functions, but I can change `processCommit` function for each batch type to accept additional `postStateRoot` as argument and return only `Types.Result` instead of postStateRoot and result.
What do you think? @jacque006

### DIFF
--- a/contracts/Transfer.sol
+++ b/contracts/Transfer.sol
@@ -23,17 +23,17 @@ contract Transfer {
      * @notice processes the state transition of a commitment
      * */
     function processTransferCommit(
-        bytes32 stateRoot,
+        bytes32 postStateRoot,
+        bytes32 previousStateRoot,
         uint256 maxTxSize,
         uint256 feeReceiver,
         bytes memory txs,
         Types.StateMerkleProof[] memory proofs
-    ) public pure returns (bytes32, Types.Result result) {
-        if (txs.transferHasExcessData())
-            return (stateRoot, Types.Result.BadCompression);
+    ) public pure returns (Types.Result result) {
+        if (txs.transferHasExcessData()) return Types.Result.BadCompression;
 
         uint256 size = txs.transferSize();
-        if (size > maxTxSize) return (stateRoot, Types.Result.TooManyTx);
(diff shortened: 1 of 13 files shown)


jacque006:
Definitely like this change, improves clarity of rollback reason in that case.

> but I can change processCommit function for each batch type to accept additional postStateRoot as argument and return only Types.Result instead of postStateRoot and result.

^ I think this is the way to go to encapsulate the validation logic in `process*Commit` functions. However, I would double check the gas cost difference to make sure it doesn't change that significantly.

And finally, if it's feasible test cases for this new Result state would be appreciated :)

b-tarczynski:
I changed `process*Commit` functions to be responsible for validation. Gas cost of `disputeTransition*` functions are the same as before changes.

contracts/Transfer.sol
```diff
@@ -23,17 +23,17 @@ contract Transfer {
      * @notice processes the state transition of a commitment
      * */
     function processTransferCommit(
-        bytes32 stateRoot,
+        bytes32 postStateRoot,
+        bytes32 previousStateRoot,
```
msieczko, line 27: I'd call this param `currentStateRoot` or `stateRoot` here to better reflect the meaning of the variable during the function execution.

contracts/test/TestTransfer.sol
```diff
@@ -40,21 +40,23 @@ contract TestTransfer is Transfer {
     }
 
     function testProcessTransferCommit(
-        bytes32 stateRoot,
+        bytes32 postStateRoot,
+        bytes32 previousStateRoot,
```
msieczko, line 44:
```suggestion
        bytes32 postStateRoot,
        bytes32 preStateRoot,
```
`preStateRoot` looks nicer next to `postStateRoot` to me. Also I'd put _pre_ before _post_.

msieczko (approved):
Overall LGTM. I only have some minor style comments.

jacque006 (approved):

Issue #664, opened 2021-09-14
Withdraw roots colliding


Marik-D:
It look like WithdrawManager stores pending withdrawals using `withdrawRoot` which is the root of the merkle tree formed by new user states generated after mass-migration batch.

The states are formed without a nonce:

```solidity
    Types.UserState memory state =
            Types.UserState({
                pubkeyID: pubkeyID,
                tokenID: tokenID,
                balance: amount,
                nonce: 0
            });
```

In that case, does this allow for a batch to be forged in a way that will cause a collision in `withdrawRoot`?

https://github.dev/thehubbleproject/hubble-contracts/blob/7058cba1a4e5a6251571f74a6a7f16de5b77e2a7/contracts/libs/Transition.sol#L217-L217

https://github.dev/thehubbleproject/hubble-contracts/blob/7058cba1a4e5a6251571f74a6a7f16de5b77e2a7/contracts/WithdrawManager.sol#L47-L47

Marik-D:
The simplest case where such a collision might happen is two MM commitments, each with a single transaction. In that case withdraw root for each of them would be calculated with:

```
withdrawRoot = keccak256(Types.UserState({
    pubkeyID: pubkeyID,
    tokenID: tokenID,
    balance: amount,
    nonce: 0
}).encode());
```

Now, as long as pubkey id, token id, and amount are the same in both transactions, which is a valid case, then the withdraw roots would be the same for both commitments.

PR #679, opened 2022-01-04
Disallow calling processWithdrawCommitment twice

duckception:
Currently, anyone can call `processWithdrawCommitment` multiple times which may lead to permanent lock up of tokens in the `WithdrawManager`. This PR fixes it and removes redundant code from the `Vault` smart contract.

### DIFF
--- a/contracts/client/FrontendMassMigration.sol
+++ b/contracts/client/FrontendMassMigration.sol
@@ -123,7 +123,8 @@ contract FrontendMassMigration {
         newReceiver = Transition.createState(
             sender.pubkeyID,
             tokenID,
-            _tx.amount
+            _tx.amount,
+            sender.nonce
         );
         return (sender.encode(), newReceiver, result);
     }
--- a/contracts/client/FrontendMassMigration.sol
+++ b/contracts/client/FrontendMassMigration.sol
@@ -120,7 +120,8 @@ contract FrontendMassMigration {
             sender
         );
         if (result != Types.Result.Ok) return (sender.encode(), "", result);
-        newReceiver = Transition.createState(
+        newReceiver = Transition.createMMState(
+            _tx.fromIndex,
             sender.pubkeyID,
             tokenID,
             _tx.amount,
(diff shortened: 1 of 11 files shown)


jacque006:
I agree with @msieczko , resolving https://github.com/thehubbleproject/hubble-contracts/issues/664 would be a better path forward. @duckception Appreciate you taking a stab at trying to resolve this.

duckception:
Issue described in #664 should be now resolved 🚀 

Additionally, I have a question @jacque006. What is the purpose of `validateAndApply` functions in `FrontendCreate2Transfer.sol` and `FrontendMassMigration.sol` contracts?

duckception:
We've discovered that changing just the nonce is not enough to prevent withdraw roots collision. Hubble allows registration of multiple user states for the same pubkey ID, therefore there still can be a case where withdraw root collision could happen (see modified test in `rollup.massMigration.test.ts`). To overcome this, I've added a new user state structure, just for mass migrations, that additionally takes into account state ID of the sender and is now used to generate collision-free withdraw root.

contracts/client/FrontendMassMigration.sol
```diff
@@ -123,7 +123,8 @@ contract FrontendMassMigration {
         newReceiver = Transition.createState(
             sender.pubkeyID,
             tokenID,
-            _tx.amount
+            _tx.amount,
+            sender.nonce
         );
```
msieczko: Use `Transition.createMMState` here for consistency

contracts/libs/Transition.sol
```diff
@@ -212,14 +219,33 @@ library Transition {
     function createState(
         uint256 pubkeyID,
         uint256 tokenID,
-        uint256 amount
+        uint256 amount,
+        uint256 nonce
```
msieczko: After introducing `createMMState` this additional parameter is unnecessary and will always be set to `0`

msieczko (changes requested):

msieczko (approved):
Everything looks good now! @jacque006, could you take a look? I think this PR is ready to be merged. We should also merge #684 right after this one.

